PR #2, opened 2023-02-02
chore: added ESLint as part of PR jobs

IgnatG:
### Changelog
- added ESLint as a required step on PR jobs

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,9 +14,5 @@ jobs:
         cache: 'npm'
     - name: Install modules
       run: npm ci
-    - name: Run Hadolint (Dockerfile)
-      uses: hadolint/hadolint-action@v3.0.0
-      with:
-        dockerfile: Dockerfile
     - name: Run ESLint (TypeScript)
       run: npm run lint:fix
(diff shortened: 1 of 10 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '18'
+        cache: 'npm'
+    - name: Install modules
+      run: npm ci
+    - name: Run ESLint (TypeScript)
+      run: npm run lint:fix
```
abhi-markan, line 18: Should this not be `npm run lint` `fixing` lint changes in the pipeline will not be written to the repo.

abhi-markan (commented):
Do we also need `.eslintrc.json`?

.github/workflows/lint.yml
IgnatG: that's true. I was testing it to see if it changes is in the pipeline - but it doesn't. So I'm taking it out

IgnatG:
> Do we also need `.eslintrc.json`?

we have `.eslintrc.js`

abhi-markan (approved):

PR #8, opened 2023-02-08
feat: added `pinojs` logging

IgnatG:
#### Changelog
- added `pinojs` logging

**Note:** writing logs should be done using:

```bash
# error
this.logger.error({ id: 'your message here' }, 'context-name');

# log
this.logger.log({ id: 'your message here' }, 'context-name');

```


avaitonis (approved):
LGTM

PR #10, opened 2023-02-10
feat: added tests for GET `/markets`

avaitonis:
* E2e test for GET /markets
* Added name for database factories, this helps to app.close() after test suite is complete
* Expected response payloads are in separate files
* Test is expecting just 9 test /markets to be in the test DB
* Change values for query parameter "active" to match Mulesoft implementation. Added validation of possible values.
* TODO: eslint gives warning for using enum. Multiple solutions exist, current one disables eslint check before enum definition.
* Removed import package.json from app.config.ts, because it was failing in partial app initialization. Maybe different initialization will be required in future.


e2e/jest-e2e.json
```diff
@@ -0,0 +1,14 @@
+{
+  "moduleFileExtensions": [
```
IgnatG, line 2: is this file being used?

test/app.e2e-spec.ts
```diff
@@ -1,7 +1,7 @@
 import { Test, TestingModule } from '@nestjs/testing';
```
IgnatG, line 1: the `e2e` naming convention for tests is associated to UI testing (i.e. testing the end to end journey of an application). For our case, it feels like it's an api test overall. so the file name can be `markets.spec.ts` (without the `e2e` part)

e2e/jest-e2e.json
```diff
@@ -0,0 +1,14 @@
...
+      "ts",
```
IgnatG, line 3: we already have the `jest.config.js` file. I'm not sure this file is needed though

package.json
```diff
@@ -18,6 +18,10 @@
     "lint": "eslint ./src --ext .ts",
     "lint:fix": "eslint ./src --ext .ts --fix",
     "test": "jest",
+    "test:watch": "jest --watch",
+    "test:cov": "jest --coverage",
+    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "jest --config ./test/jest-e2e.json",
```
IgnatG, line 24: we can move the `jest.config.js` file into `test/` folder. alternatively, we can leave it where it is an simply use that instead (in which case, we can delete the `jest-e2e.json` files)

test/markets/markets.e2e-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  beforeAll(async () => {
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [MainModule],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+
+    // Validation pipeline is require to check validations.
+    app.useGlobalPipes(
```
IgnatG, line 19: I'm getting the impression that this is something we can re-use for all tests. In which case, can we create a separate file that has the boilerplate initialization for these tests? and then we import it where we need it

test/markets/markets.e2e-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        transformOptions: {
+          enableImplicitConversion: true,
+        },
+      }),
+    );
+    await app.init();
+  });
+
+  it(`GET /markets`, () => {
+    return request(app.getHttpServer()).get('/markets')
```
IgnatG, line 33: have a look at DTFS set up. we can have a separate config for `supertest` calls.

test/markets/markets.e2e-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      .expect(200)
+      .expect('Content-Type', 'application/json; charset=utf-8')
+      .expect(require('./expected-responses/GET-markets.json'));
```
IgnatG, line 36: 1. can we import the file at the top and assign it to a variable? 
2. the `expect` part should be something like `expect(response)`.toEqual(get-markets-expectation);` - see this page https://jestjs.io/docs/expect
3. can we execute the `expect` statements on separate lines? i.e. 
`expect(response.body)`.toEqual(get-markets-expectation);`
`expect(response.status)`.toEqual({status: 200});`
or something like that ^

test/markets/markets.e2e-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+      }),
+    );
+    await app.init();
+  });
+
+  it(`GET /markets`, () => {
+    return request(app.getHttpServer()).get('/markets')
+      .expect(200)
+      .expect('Content-Type', 'application/json; charset=utf-8')
```
IgnatG, line 35: this one should not be needed

test/markets/markets.e2e-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  it(`GET /markets?active=something-else`, () => {
+    return request(app.getHttpServer()).get('/markets?active=something-else')
+      .expect(400)
+      .expect('Content-Type', 'application/json; charset=utf-8')
+      .then(response => {
+        expect(response.body.error).toMatch('Bad Request');
+        expect(response.body.message[0]).toMatch('active must be one of the following values: Y, N');
+      });
+  });
```
IgnatG, line 61: can we other scenarions for:
1. `/markets?active=`
2. `/markets`
3. `/markets?active=undefined`
4. `/markets?active=null`

src/modules/markets/markets.service.ts
```diff
@@ -15,7 +15,7 @@ export class MarketsService {
     let spResults = await this.marketsRepository.query('CIS_USP_READ_MARKETS');
```
IgnatG, line 15: just a thought, should/could this be a table? rather than SP?

IgnatG: because in that case we could use something like
```
    let filter = {};
    if (active === 'Y') {
      filter = { active: 'Y' };
    } else if (active === 'N') {
      filter = { active: 'N' };
    }

    const response = await this.marketsRepository.find({ where: [filter] });
```

IgnatG, line 15: we should wrap the entire content of this function in a try-catch statement. This so that it doesn't crash the server if something goes sideways

test/app.e2e-spec.ts
avaitonis: *.e2e-spec.ts naming convention is used in NestJs samples. *.spec.ts is used for unit tests, but these files are in different locations so we could use same suffix.
Should I still change to *.spec.ts?

IgnatG: ok. let's use `*.api-test.ts` in that case. most likely we will need unit tests as well at some point

package.json
avaitonis: Top level jest file is for Unit tests. We don't have them at the moment, but they should be close to class files and not in /test folder. 

In new nestJs samples jest.config.js is named as jest.json and in json format.

IgnatG: ok cool. can we use the same naming convention as above though (i.e. `jest.api-test.js/json`? we can use either `*.js` or `*.json` - we just need to specify where the config file is in package.json

e2e/jest-e2e.json
avaitonis: yes, this file and e2e folder were committed by mistake. Removed

package.json
avaitonis: Renamed in this repo. We can also rename e2e-spec to api-test in nestjs-template

test/markets/markets.e2e-spec.ts
avaitonis: Moved app initialization to separate file. It is still not reusing main initialisation, couldn't get it working. NestJs bootstrap is slightly different from DTFS API bootstrap.

test/markets/markets.e2e-spec.ts
avaitonis: Copied api.js from DTFS. Because of app initialisation differences api.js is loaded not at the top of test file.

test/markets/markets.e2e-spec.ts
avaitonis: 1. Done
2. NestJs mostly uses chained checks in samples and own tests, but I unchained them.
3. Done

test/markets/markets.e2e-spec.ts
avaitonis: Removed

test/markets/markets.e2e-spec.ts
avaitonis: Added. All input validation checks are returning same error, so it is some test code duplication. Not sure if it should be optimised somehow.

src/modules/markets/markets.service.ts
avaitonis: Added try catch. User gets generic error message and all details are logged. Maybe global exception handler can be added later.

test/createApp.ts
```diff
@@ -0,0 +1,34 @@
+import { INestApplication, ValidationPipe } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import { MainModule } from '../src/main.module';
+
+
+export class CreateApp {
+  // NestJs app initialisation happens in /src/main.ts, here is some duplication of same process.
+  // TODO: maybe it is possible to reuse initialisation in /src/main.ts as similar approach is done in DTFS projects.
```
IgnatG, line 8: agreed. but for now it can stay as it is

PR #11, opened 2023-02-12
feat: added `cspell`, custom paths, lint staged files & updated all npm packages

IgnatG:
### Changelog
- added `lint-staged` (this will lint staged files)
- order packages inside `package.json` automatically, via `sort-package-json`
- updated all npm packages
- added cspell (to check for spelling mistakes)
- updated the linting rules and converted the `.js` file to `.json` (to keep the file names consistent)
- added custom paths via `tsconfig.json`


src/modules/markets/entities/market.entity.ts
```diff
@@ -59,11 +59,11 @@ export class MarketEntity {
 
   @Column({ name: 'ESRA_ID' })
   @ApiProperty({ example: 1, description: '' })
-  ESRAClasificationId: number;
+  ESRAClassificationId: number;
```
IgnatG, line 62: @avaitonis  can you check if this is correct? there was a typo here

src/modules/markets/entities/market.entity.ts
```diff
@@ -59,11 +59,11 @@ export class MarketEntity {
...
 
   @Column({ name: 'ESRA_DESC' })
   @ApiProperty({ example: 'Standard Risk', description: '' })
-  ESRAClasificationDesc: string;
+  ESRAClassificationDesc: string;
```
IgnatG, line 66: same as above, @avaitonis  can you check if this is correct? there a typo here

abhi-markan (approved):

src/modules/markets/entities/market.entity.ts
avaitonis: This typo was introduced 7 months ago and endpoint is now used by EXIP. We can fix typo but release of this nodeJS API will require small change in EXIP project.

src/modules/markets/entities/market.entity.ts
avaitonis: This typo was introduced 7 months ago and endpoint is now used by EXIP. We can fix typo but release of this nodeJS API will require small change in EXIP project.

avaitonis (approved):

src/modules/markets/entities/market.entity.ts
IgnatG: ok cool. let's fix it as part of this PR and we will notify Tony once the API is live

PR #17, opened 2023-02-17
feat(API-69): added `/interest-rates` endpoint

IgnatG:
### Changelog
- added GET `/interest-rate` endpoint
- updated the TSConfig file to include test files (i.e. anything that's inside `/test` folder)


src/modules/interest-rates/entities/interest-rate.entity.ts
```diff
@@ -0,0 +1,58 @@
+import { ClassSerializerInterceptor, UseInterceptors } from '@nestjs/common';
+import { ApiProperty } from '@nestjs/swagger';
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity({
+  name: 'DWR_INT_RATE',
+  database: 'cedar',
```
avaitonis, line 7: database name is not required and it works better without if we use one DB for testing.

src/modules/interest-rates/interest-rates.service.ts
```diff
@@ -0,0 +1,22 @@
...
+@Injectable()
+export class InterestRatesService {
+  constructor(
+    @InjectRepository(InterestRatesEntity, 'mssql-cis')
+    private readonly interestRates: Repository<InterestRatesEntity>,
+  ) {}
+
+  async findAll() {
+    try {
+      const response = await this.interestRates.find();
```
avaitonis, line 16: DWR_INT_RATE table has history or all rates. In DEV environment we have 56462 records, not sure how this be handled here.

To return just active rates we need to filter with column EFFECTIVE_TO_DATETIME = '99991231' , not sure yet how to add it.

IgnatG: good shout - I've added the filter now

src/modules/interest-rates/entities/interest-rate.entity.ts
IgnatG: I've tried to remove it but for some reason, it breaks the server

test/api.ts
```diff
@@ -35,33 +32,13 @@ export class Api {
     };
   }
 
-  put(data) {
+  put(data: string | object) {
     return { to: (url: string) => request(this.app).put(url).send(data) };
   }
 
-  putMultipartForm(data: FormData[], files = []) {
```
IgnatG, line 42: this is only needed if we have forms on the page; it's not applicable for us

test/markets/markets.api-test.ts
```diff
@@ -17,46 +17,46 @@ describe('Markets', () => {
 
   it(`GET /markets`, async () => {
     const { status, body } = await api.get('/markets');
-    expect(status).toEqual(200);
+    expect(status).toBe(200);
```
IgnatG, line 20: these changes are mainly driven by ESLint - it was disabled previously for test files

src/modules/interest-rates/entities/interest-rate.entity.ts
IgnatG: updated

src/modules/interest-rates/interest-rates.service.ts
```diff
@@ -0,0 +1,23 @@
...
+export class InterestRatesService {
+  private readonly logger = new Logger();
+  constructor(
+    @InjectRepository(InterestRatesEntity, 'mssql-cedar')
+    private readonly interestRates: Repository<InterestRatesEntity>,
+  ) {}
+
+  findAll(): Promise<InterestRatesEntity[]> {
+    try {
+      return this.interestRates.find({ where: { effectiveTo: Equal('9999-12-31 00:00:00.000') } });
```
IgnatG, line 17: @avaitonis is this filter what you'd expect here? i.e. return everything that has effectiveTo set to `'9999-12-31 00:00:00.000'`

test/api.ts
avaitonis: Makes sense, happy to remove it now or later

src/modules/interest-rates/interest-rates.service.ts
avaitonis: Looks good. I assume it works. It might also work without time part, just "9999-12-31".

avaitonis (approved):
LGTM

PR #20, opened 2023-02-20
feat: ensure new ids are in order

avaitonis:
## Introduction

* Fix entities/ukef-id.entity.ts field type
* Ensure new Ids are in order
* e2e/api Test that POST /numbers IDs are in order
* Unit tests for new helper functions

##  Miscellaneous
* @abhi-markan Added `npm ci` to `lint.yml`.


src/modules/numbers/numbers.service.ts
```diff
@@ -64,4 +66,41 @@ export class NumbersService {
...
+    return Object.entries(newIdsByType).reduce((acc, typeWithIds) => {
+      acc[typeWithIds[0]] = typeWithIds[1].sort();
+      return acc;
+    }, Object.create(null));
+  }
+
+  /**
+   * Helper to group ids by type, so they can be sorted.
+   */
+  groupIdsByType(newUkefIds: UkefId[]): unknown {
```
IgnatG, line 99: Should this be moved to a helper function? we will also need a unit test for it to make sure it works as expected

src/modules/numbers/numbers.service.ts
```diff
@@ -64,4 +66,41 @@ export class NumbersService {
...
+      newId['maskedId'] = sortednewIdsByType[newId['type']].shift();
+      sortedIds.push(newId);
+    });
+    return sortedIds;
+  }
+
+  /**
+   * Helper to sort new Ids inside type array.
+   */
+  sortGroupedByTypeIds(newIdsByType: unknown): unknown {
```
IgnatG, line 89: Should this be moved to a helper function? we will also need a unit test for it to make sure it works as expected

src/modules/numbers/numbers.service.ts
```diff
@@ -64,4 +66,41 @@ export class NumbersService {
       requestingSystem: dbNumber.REQUESTING_SYSTEM,
     };
   }
+
+  /**
+   * Helper to sort new ids
+   */
+  sortIds(newUkefIds: UkefId[]): UkefId[] {
```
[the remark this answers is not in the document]
IgnatG: Never mind, just seen the tests now - all good on that front

src/modules/numbers/numbers.service.ts
IgnatG: Never mind, just seen the tests now - all good on that front

src/modules/numbers/numbers.service.ts
IgnatG: Never mind, just saw the tests now - all good on that front

src/modules/numbers/numbers.service.ts
```diff
@@ -35,7 +37,7 @@ export class NumbersService {
       if (!dbNumber[0]) {
```
abhi-markan, line 37: Can this please be replaced with `dbNumber && !dbNumber[0]`, in case `dbNumber` is not an array.

src/modules/numbers/numbers.service.ts
```diff
@@ -64,4 +66,41 @@ export class NumbersService {
...
     };
   }
+
+  /**
+   * Helper to sort new ids
+   */
+  sortIds(newUkefIds: UkefId[]): UkefId[] {
+    const sortedIds = [];
+    const newIdsByType = this.groupIdsByType(newUkefIds);
+    const sortednewIdsByType = this.sortGroupedByTypeIds(newIdsByType);
```
abhi-markan, line 76: Will `newIdsByType` ever be `null`, `undefined`? If yes we will need to add void check before any further execution.

abhi-markan, line 76: Same for `sortednewIdsByType` please.

abhi-markan (changes requested):
Few comments to address please.

src/modules/numbers/numbers.service.ts
avaitonis: I am not expecting to reuse this function and service class looks like a good place for it. Do we still want to move it out?

src/modules/numbers/numbers.service.ts
avaitonis: I am not expecting to reuse this function and service class looks like a good place for it. Do we still want to move it out?

src/modules/numbers/numbers.service.ts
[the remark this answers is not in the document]
avaitonis: I am not expecting to reuse this function and service class looks like a good place for it. Do we still want to move it out?

src/modules/numbers/numbers.service.ts
IgnatG: Ok. Let's leave it as it is for now

src/modules/numbers/numbers.service.ts
IgnatG: Ok. Let's leave it as it is for now

src/modules/numbers/numbers.service.ts
avaitonis: Hi,
Can't find a way how newIdsByType would be null or undefined. Input is validated and DB responses checked before this. So maybe check is not required.

avaitonis: Hi,
Can't find a way how newIdsByType would be null or undefined. Input is validated and DB responses checked before this. So maybe check is not required.

src/modules/numbers/numbers.service.ts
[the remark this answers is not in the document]
IgnatG: nope - leave it as it is for now

src/modules/numbers/numbers.service.ts
avaitonis: We agreed that checks in above code is enough.

src/modules/numbers/numbers.service.ts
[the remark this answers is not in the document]
avaitonis: Might refactor later, as in comment above.

abhi-markan (approved):

PR #21, opened 2023-02-20
feat(API-82): Improved GHA

abhi-markan:
## Introduction
Purpose of this PR is to enhance existing GHA, introduction of new GHA and also to bring them inline with the rest of UKEF GHA.

## Resolution
* Consistent `lint.yml`
* Added `deploy.sh` script for deployment.
* Consistent auto release GHA.
* Enhanced SCA.
* Enabled dependency review at GH level and removed `dependency-review.yml` GHA.


.github/workflows/lint.yml
```diff
@@ -1,18 +1,59 @@
...
\ No newline at end of file
+    - name: Environment 🧪
+      run: echo "Environment set to ${{ env.environment }}"
+  
+    - name: Timezone 🌐
+      run: echo "Timezone set to ${{ env.timezone }}"
+
+# 2. Lint
+  lint:
+    name: Scanning 🎨
```
IgnatG, line 38: should the name be "Linting"?

abhi-markan: It only scans for lint error, no linting is performed at GHA level.

IgnatG: yeah, so something like "Lint scanning" - otherwise it can be security scanning, or something different. 
this so that we avoid ambiguity

abhi-markan: There will be no ambiguity as the Job name will be displayed in conjuction with the step name, so it will appear like this.

image

IgnatG (approved):

PR #25, opened 2023-02-21
fix: remove GET /number-types endpoint, its not used by clients

avaitonis:


IgnatG (approved):
LGTM - just make sure you address the Linting errors as well

IgnatG:
> LGTM - just make sure you address the Linting errors as well

actually, never mind. it's a configuration error that can be solved as part of a different PR

PR #27, opened 2023-02-21
feat: make markets API tests more generic

avaitonis:
- make markets API tests more generic
- fix GET ukefId API test for 404 case


IgnatG (approved):
LGTM

PR #30, opened 2023-02-22
feat: added GET `/sector-industries` endpoint

avaitonis:
GET /sector-industries
old endpoint name was GET /map-industry-sector
text trimming is done in entity class using transformer attribute.


IgnatG (dismissed):
LGTM overall - I left a small comment; not a blocker though

src/modules/sector-industries/sector-industries.controller.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import { SectorIndustriesController } from './sector-industries.controller';
+import { SectorIndustriesService } from './sector-industries.service';
+
+const chance = new Chance();
+
+describe('SectorIndustriesController', () => {
+  let sectorIndustriesController: SectorIndustriesController;
+  let sectorIndustriesService: SectorIndustriesService;
+
+  beforeEach(async () => {
```
abhi-markan, line 14: Can this please be `before` instead of `beforeEach`?

src/modules/sector-industries/sector-industries.controller.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    sectorIndustriesController = app.get<SectorIndustriesController>(SectorIndustriesController);
+    sectorIndustriesService = app.get<SectorIndustriesService>(SectorIndustriesService);
+  });
+
+  it('should be defined', () => {
+    expect(sectorIndustriesController).toBeDefined();
+  });
+
+  describe('find()', () => {
+    it('should return all sector industry records', () => {
```
abhi-markan, line 54: Can we add following test cases please:

* Calling a specific industry.
* Return response is `null` or not `200/201`?

src/modules/sector-industries/sector-industries.controller.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      const query = new GetSectorIndustriesQueryDto();
+      sectorIndustriesController.find(query);
+
+      expect(sectorIndustriesService.find).toHaveBeenCalled();
+    });
+  });
+});
```
abhi-markan, line 61: With the latest PR for test methodology consistency, can this file be renamed from `sector-industries.controller.spec.ts` to `sector-industries.controller.test.ts`. Thanks

src/modules/sector-industries/sector-industries.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  constructor(
+    @InjectRepository(SectorIndustryEntity, 'mssql-mdm')
+    private readonly sectorIndustries: Repository<SectorIndustryEntity>,
+  ) {}
+
+  async find(ukefSectorIdInput, ukefIndustryId): Promise<SectorIndustryEntity[]> {
+    try {
+      let whereConditions: object;
+
+      whereConditions = { effectiveTo: Equal(new Date('9999-12-31 00:00:00.000')) };
```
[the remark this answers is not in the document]
abhi-markan: Can the above value be consumed from a centralised constant please? Happy to create a PR with centralised constants.

src/modules/sector-industries/sector-industries.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      const results = await this.sectorIndustries.find({ where: whereConditions });
+
+      if (results && !results[0]) {
+        throw new NotFoundException('No results for your search criteria');
+      }
+
+      return results;
+    } catch (err) {
+      if (err instanceof NotFoundException) {
+        throw err;
```
abhi-markan, line 36: Do we need to add `logger.error` here so it is displayed on the terminal or will it be displayed by the calling stack exceptional handler?

abhi-markan (changes requested):
Good work, few comments to address please.

src/modules/sector-industries/sector-industries.service.ts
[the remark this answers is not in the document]
avaitonis: Renamed and setting default value in one line.

Centralised constants would be great, lets do in separate PR for all modules.

src/modules/sector-industries/sector-industries.controller.spec.ts
avaitonis: Good catch, I think you changed beforeEach to beforeAll.

src/modules/sector-industries/sector-industries.controller.spec.ts
avaitonis: Renamed

src/modules/sector-industries/sector-industries.service.ts
avaitonis: I added logger.warn as this is not very important error and we shouldn't get alerts.

Other logging level could be more appropriate than logger.warn. We could change to logger.debug, logger.log, logger.verbose. Maybe do this logging in global error handler.

src/modules/sector-industries/sector-industries.controller.spec.ts
avaitonis: We have 404 test bellow, it is "GET /sector-industries?ukefSectorId=1234" , 1234 is not existing sector.

Maybe test type (404) should be more obvious.

src/modules/sector-industries/sector-industries.controller.test.ts
```diff
@@ -0,0 +1,61 @@
...
+import { SectorIndustriesController } from './sector-industries.controller';
+import { SectorIndustriesService } from './sector-industries.service';
+
+const chance = new Chance();
+
+describe('SectorIndustriesController', () => {
+  let sectorIndustriesController: SectorIndustriesController;
+  let sectorIndustriesService: SectorIndustriesService;
+
+  beforeEach(async () => {
```
abhi-markan, line 14: Sorry can this `beforeEach` please be updated to `before` or `beforeAll`.

abhi-markan (changes requested):
One minor comment to address please.

src/modules/sector-industries/sector-industries.controller.test.ts
avaitonis: Changed to beforeAll

abhi-markan (approved):

PR #38, opened 2023-02-24
feat(API-83): GET `/constants/spi`

avaitonis:
## Introduction
New endpoint GET /constants/spi returns financial constants required for SPI website to calculate indicative sovereign premium rate

## Resolution
New module constants, it might be expanded later for other similar endpoints other than /constants/spi
Using SP USP_MDM_READ_SPI_CONSTANTS @0, @1 with 2 optional parameters
Result is in slightly different format from current/old endpoint, but it is compatible.

## Miscellaneous
Swagger documentation improvements in other modules


src/modules/constants/constants.service.ts
```diff
@@ -0,0 +1,44 @@
+import { Injectable, InternalServerErrorException, Logger, NotFoundException } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+import { DbResponseHelper } from '../../helpers/db-response.helper';
```
IgnatG, line 5: should this come from an alias path? i.e. `@ukef/helpers/...`

IgnatG (dismissed):
LGTM overall - great work!

src/modules/constants/constants.service.ts
avaitonis: Updated

IgnatG (approved):

PR #39, opened 2023-02-24
feat: new endpoint GET /exposure-period

avaitonis:
### Introduction
New endpoint GET /exposure-period. It calculates UKEF exposure in months for EW and BS Facilities

###  Resolution
GET /exposure-period uses SP USP_MDM_READ_EXPOSURE_PERIOD 
Didn't use TypeORM repository and entity because it is no table behind SP
No DB tables are used for calculation, so in future SP code could be migrated to JavaScript

### Miscellaneous
Constants API tests improvement

### DIFF
--- a/src/modules/exposure-period/exposure-period.service.ts
+++ b/src/modules/exposure-period/exposure-period.service.ts
@@ -1,0 +1,31 @@
+import { Injectable, InternalServerErrorException, Logger } from '@nestjs/common';
+import { InjectDataSource } from '@nestjs/typeorm';
+import { DataSource } from 'typeorm';
+
+import { ExposurePeriodDto } from './dto/exposure-period.dto';
+
+@Injectable()
+export class ExposurePeriodService {
+  private readonly logger = new Logger();
+
+  constructor(
+    @InjectDataSource('mssql-mdm')
+    private readonly mdmDataSource: DataSource,
+  ) {}
+
+  async calculate(startDate: Date, endDate: Date, productGroup: string): Promise<ExposurePeriodDto> {
+    try {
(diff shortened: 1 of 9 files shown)


src/modules/exposure-period/exposure-period.service.ts
```diff
@@ -0,0 +1,31 @@
...
+    private readonly mdmDataSource: DataSource,
+  ) {}
+
+  async calculate(startDate: Date, endDate: Date, productGroup: string): Promise<ExposurePeriodDto> {
+    try {
+      // TODO: SP USP_MDM_READ_EXPOSURE_PERIOD is not using data/tables from DB. Calculation could be moved to Javascript.
+      const spResults = await this.mdmDataSource.query('USP_MDM_READ_EXPOSURE_PERIOD @0, @1, @2', [startDate, endDate, productGroup]);
+
+      if (!spResults || !spResults[0] || typeof spResults[0].EXPOSURE_PERIOD === 'undefined') {
+        throw new InternalServerErrorException('No exposure period result from USP_MDM_READ_EXPOSURE_PERIOD');
```
IgnatG, line 22: should this be a 404 instead? rather than 500 type of error?

avaitonis: If input is not correct it is 400.
If input is correct it should always return calculation. Simplified calculation is endDate - startDate in months (rounded up to full month). 
I think 500 is better code in case calculation doesn't have result and something is wrong on server side.

IgnatG: makes sense

IgnatG (approved):
LGTM

PR #44, opened 2023-03-05
feat: added authentication using API Key

IgnatG:
### Changelog
- Added authentication using API Key
- small code refactoring for API tests


avaitonis (approved):
LGTM

src/auth/auth.module.ts
```diff
@@ -0,0 +1,12 @@
+import { Module } from '@nestjs/common';
```
lmarrai-sw, line 1: Is there a reason the auth module is not under `/modules/auth`?

src/auth/auth.service.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  it('should be defined', () => {
+    expect(spyService).toBeDefined();
+  });
+
+  it('should call `validateApiKey` with expected param', () => {
+    const key = chance.word();
+    spyService.validateApiKey(key);
+
+    expect(spyService.validateApiKey).toHaveBeenCalledWith(key);
+  });
```
lmarrai-sw, line 41: I don't think this test is quite right: line 40 will always pass because of line 38, we're not testing any code within the real `AuthService`

I expected instead that the test would:

1. mock `ConfigService` to return a particular value for `'app.apiKey'`
2. call `authService.validateApiKey` with a real auth service
3. check that it returns true if the input matches the mocked value from step 1, or false if it does not

src/auth/auth.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Injectable } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+
+@Injectable()
+export class AuthService {
+  constructor(private configService: ConfigService) {}
```
lmarrai-sw, line 6: we can make the `configService` `readonly` as well

src/auth/auth.service.ts
```diff
@@ -0,0 +1,12 @@
...
+
+  validateApiKey(key: string) {
```
lmarrai-sw, line 8: We can specify a return type on this function

src/auth/auth.module.ts
IgnatG: in reality, it can be anywhere
We tend to have at the top level items that are globally used - i.e. `/config`, `/database`, /helpers` etc
Although this can easily be as part of the `modules/` folder, so happy to move it if needed

src/database/database.module.ts
```diff
@@ -0,0 +1,12 @@
...
+import { MsSqlCedarDatabaseModule } from './mssql-cedar-database.module';
+import { MsSqlCisDatabaseModule } from './mssql-cis-database.module';
+import { MsSqlMdmDatabaseModule } from './mssql-mdm-database.module';
+import { MsSqlNumberGeneratorDatabaseModule } from './mssql-number-generator-database.module';
+
+@Module({
+  imports: [MsSqlMdmDatabaseModule, MsSqlCedarDatabaseModule, MsSqlCisDatabaseModule, MsSqlNumberGeneratorDatabaseModule],
+  exports: [MsSqlMdmDatabaseModule, MsSqlCedarDatabaseModule, MsSqlCisDatabaseModule, MsSqlNumberGeneratorDatabaseModule],
+})
+export class DatabaseModule {}
```
lmarrai-sw, line 12: Now that we have a `DatabaseModule`, is there any reason why the `database` folder isn't under `module`?

src/modules/constants/constants.controller.ts
```diff
@@ -4,7 +4,6 @@ import { ApiBearerAuth, ApiOperation, ApiParam, ApiResponse, ApiTags } from '@ne
 import { ConstantsService } from './constants.service';
 import { GetConstantsSpiQueryDto } from './dto/get-constants-spi-query.dto';
 import { ConstantSpiEntity } from './entities/constants-spi.entity';
-
 @ApiBearerAuth()
```
lmarrai-sw, line 7: I'm surprised we don't have a blank line between the last import and our code here

src/auth/strategy/apikey.strategy.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { PassportStrategy } from '@nestjs/passport';
+import { HeaderAPIKeyStrategy } from 'passport-headerapikey';
+
+import { AuthService } from '../auth.service';
+
+@Injectable()
+export class ApiKeyStrategy extends PassportStrategy(HeaderAPIKeyStrategy, 'x-api-key') {
+  constructor(private authService: AuthService, private configService: ConfigService) {
```
lmarrai-sw, line 10: we can make the `authService` and `configService` readonly

src/auth/strategy/apikey.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+    const headerKeyApiKey: string = configService.get<string>('app.apiKeyStrategy');
+    super({ header: headerKeyApiKey, prefix: '' }, true, (apiKey: string, done: (arg0: UnauthorizedException, arg1: boolean) => void) => {
+      const hasValidKey: boolean = this.authService.validateApiKey(apiKey);
+      if (hasValidKey) {
+        done(null, true);
```
lmarrai-sw, line 15: I think this should be `return done(null, true);` instead, so we don't execute the 2nd `done` after this

src/main.module.ts
```diff
@@ -24,13 +26,11 @@ import { LoggerModule } from 'nestjs-pino';
...
-    MsSqlCedarDatabaseModule,
-    MsSqlCisDatabaseModule,
-    MsSqlNumberGeneratorDatabaseModule,
+    AuthModule,
+    DatabaseModule,
     MdmModule,
   ],
   controllers: [],
-  providers: [],
+  providers: [AuthService],
```
lmarrai-sw, line 34: Out of curiousity, do we need to declare `AuthService` as a provider here? Is that fact that `AuthModule` declares it as a provider not enough?

src/auth/strategy/apikey.strategy.ts
```diff
@@ -0,0 +1,20 @@
...
+      }
+      done(new UnauthorizedException(), null);
+    });
```
lmarrai-sw, line 18: It would be nice to unit test this behaviour, although I imagine it would be slightly fiddly because it is in the constructor of a 3rd party class.

I think in this case we might be able to:

1. mock the HeaderAPIKeyStrategy constructor, and mock the auth service and config service
2. call the ApiKeyStrategy constructor with the mocked auth service and config service
3. test that the first 2 inputs to the HeaderAPIKeyStrategy constructor are as expected
4. get the 3rd input to the `HeaderAPIKeyStrategy` constructor (the `verify` function) and call it with an `apiKey` and mocked `done` function - then we can test that it validates the api key correctly

src/auth/strategy/apikey.strategy.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { PassportStrategy } from '@nestjs/passport';
+import { HeaderAPIKeyStrategy } from 'passport-headerapikey';
+
+import { AuthService } from '../auth.service';
+
+@Injectable()
+export class ApiKeyStrategy extends PassportStrategy(HeaderAPIKeyStrategy, 'x-api-key') {
```
lmarrai-sw, line 9: Should this file be called `api-key.strategy.ts` instead?

src/auth/guard/apiKey.guard.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class ApiKeyAuthGuard extends AuthGuard('x-api-key') {}
```
lmarrai-sw, line 5: Should this file be called `api-key.guard.ts` instead?

src/auth/auth.module.ts
lmarrai-sw: The reason I ask is that I think it's a bit surprising to have a folder called `modules` but it only has some of the `modules` in it rather than all of them. I have a preference for moving them to the `modules` folder (or renaming the `modules` folder) but if the rest of the team prefer this then that's fine with me :)

src/main.module.ts
```diff
@@ -24,13 +26,11 @@ import { LoggerModule } from 'nestjs-pino';
         },
       },
     }),
-    MsSqlMdmDatabaseModule,
-    MsSqlCedarDatabaseModule,
-    MsSqlCisDatabaseModule,
-    MsSqlNumberGeneratorDatabaseModule,
+    AuthModule,
+    DatabaseModule,
```
lmarrai-sw, line 30: Is there a reason we add these to `main.module` instead of `mdm.module`?

src/modules/constants/constants.controller.ts
IgnatG: nope - it's ESLint formatting. I'll double check

src/main.module.ts
IgnatG: good point - we don't need it here

src/auth/strategy/apikey.strategy.ts
IgnatG: yup - good point

src/auth/guard/apiKey.guard.ts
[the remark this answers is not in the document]
lmarrai-sw: Yes a `constant` :)

src/auth/auth.module.ts
avaitonis: Folder `modules` contain mdm.module.ts , so `modules` could be renamed to `mdm` or `mdm-endpoints`

src/auth/auth.service.ts
IgnatG: this is updated now

src/auth/auth.service.ts
IgnatG: this is updated now

test/markets/markets.api-test.ts
```diff
@@ -68,7 +68,7 @@ describe('Markets', () => {
     );
   });
 
-  it(`Compare GET /markets?active=N and GET /markets?active=N`, async () => {
+  it(`compares GET /markets?active=Y and GET /markets?active=N`, async () => {
```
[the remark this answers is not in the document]
IgnatG: this is updated now

src/database/database.module.ts
IgnatG: this is updated now

src/modules/constants/constants.controller.ts
IgnatG: this is updated now

src/auth/strategy/apikey.strategy.ts
IgnatG: this is updated now

src/auth/strategy/apikey.strategy.ts
IgnatG: this is updated now

src/auth/guard/apiKey.guard.ts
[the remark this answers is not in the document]
IgnatG: this is updated now

src/main.module.ts
IgnatG: this is updated now

src/auth/strategy/apikey.strategy.ts
IgnatG: this is updated now

src/auth/guard/apiKey.guard.ts
IgnatG: this is updated now

src/main.module.ts
IgnatG: this is updated now

src/auth/auth.module.ts
IgnatG: this is updated now

src/main.ts
```diff
@@ -42,6 +43,7 @@ const main = async () => {
 
   app.useLogger(app.get(Logger));
   app.useGlobalInterceptors(new TransformInterceptor());
+  app.useGlobalGuards(new ApiKeyAuthGuard());
```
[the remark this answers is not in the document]
IgnatG: agreed on Slack that it will be addressed as a separate PR

src/auth/strategy/apikey.strategy.ts
IgnatG: I'll address this once we refactor the whole test work

.lintstagedrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "*.ts": ["eslint --fix"],
+  "*.ts": ["eslint ./src --ext .ts --fix"],
```
lmarrai-sw, line 2: Do we not want to lint the `test` files as well?

package.json
```diff
@@ -13,8 +13,8 @@
     "clean": "rimraf dist/",
     "copyfiles": "copyfiles -u 1 public/* dist/public",
     "postinstall": "husky install",
-    "lint": "eslint . --ext .ts",
-    "lint:fix": "eslint . --ext .ts --fix",
+    "lint": "eslint ./src --ext .ts",
+    "lint:fix": "eslint ./src --ext .ts --fix",
```
lmarrai-sw, line 17: Same comment as [here](https://github.com/UK-Export-Finance/mdm-api/pull/44/commits/fffeedb020a499a444130a9069acaa1f9e13db67#r1128343818) - do we not want to list the `test` files as well?

src/modules/auth/auth.service.test.ts
```diff
@@ -7,36 +7,39 @@ import { AuthService } from './auth.service';
 import { ApiKeyStrategy } from './strategy/api-key.strategy';
 
 describe('AuthService', () => {
-  let spyService: AuthService;
+  let authsService: AuthService;
```
lmarrai-sw, line 10: very small nitpick, but should this be `authService` instead of `authsService`?

lmarrai-sw:
Thanks @IgnatG! I've made a few minor comments but otherwise I'm happy with the PR (I don't think I have the permissions to "approve" it in any sense)

test/markets/markets.api-test.ts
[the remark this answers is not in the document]
IgnatG: this is updated now

.lintstagedrc.json
IgnatG: this is updated now

package.json
IgnatG: this is updated now

src/modules/auth/auth.service.test.ts
IgnatG: this is updated now

package.json
IgnatG: this is updated now

avaitonis (approved):
LGTM

src/auth/auth.service.test.ts
IgnatG: this has been updated

PR #50, opened 2023-03-10
feat(APIM-96): added GET `/yield-rates` endpoint

avaitonis:
### Introduction
- New endpoint GET /yield-rates. It calculates UKEF exposure in months for EW and BS Facilities
  Query parameters: searchDate (optional)

### Resolution
- GET /exposure-period uses TypeORM to get data from table DWR_YIELD_RATE
- Old endpoint had query parameter searchDatetime, I changed it to searchDate
- Old endpoint had field short_name, I renamed to shortName

### Miscellaneous
- Removed @ApiParam from all modules because Swagger module can generate URL and Query parameter documentation from DTO class. We need URL and Query parameters in DTO for validation, so we also keep it there for documentation.
- Change MAXIMUM_TIMEZONE_LIMIT format to better match what DB response looks, and this allows to use it for testing.
- Changed GET /exposure-period Query parameter type from date to string date, it allows easier way to stop time input.


src/modules/yield-rates/yield-rates.service.ts
```diff
@@ -0,0 +1,41 @@
+import { Injectable, InternalServerErrorException, Logger, NotFoundException } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { DATABASE } from '@ukef/constants';
+import { Equal, LessThanOrEqual, MoreThan, Repository } from 'typeorm';
+
+import { DATE } from '../../constants/date.constant';
```
IgnatG, line 6: can we use the alias path?

src/modules/yield-rates/yield-rates.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    try {
+      // Search by date or use standard max date to get active/current rates.
+      const query: object = searchDate
+        ? { effectiveTo: MoreThan(searchDate), effectiveFrom: LessThanOrEqual(searchDate) }
+        : { effectiveTo: Equal(DATE.MAXIMUM_LIMIT) };
+
+      const results = await this.yieldRateRepository.find({
+        where: query,
+      });
+      if (results && !results[0]) {
```
IgnatG, line 26: results will always be an array (either empty or with values). I think we can use here:
```js
if (!results.length) {...}
```

src/modules/yield-rates/yield-rates.service.ts
avaitonis: Done, moved to line 3

src/modules/numbers/dto/get-numbers-query.dto.ts
```diff
@@ -5,12 +5,12 @@ export class GetNumbersQueryDto {
   @IsInt()
   @IsNotEmpty()
   @Max(9)
-  @ApiProperty({ example: 1 })
+  @ApiProperty({ example: 1, description: 'Id of UKEF ID type. Common types are: 1 for Deal/Facility, 2 for Party, 8 for Covenant' })
```
abhi-markan, line 8: Should we only be accepting `1, 2 or 8`?

src/modules/numbers/dto/get-numbers-query.dto.ts
```diff
@@ -5,12 +5,12 @@ export class GetNumbersQueryDto {
   @IsInt()
   @IsNotEmpty()
   @Max(9)
```
abhi-markan, line 7: Can we also add `@Min(2)`, or a number you think is a common min standard please?

src/modules/numbers/dto/get-numbers-query.dto.ts
abhi-markan: Can we also add `@Min(2)`, or a number you think is a common min standard please?

src/modules/sector-industries/dto/get-sector-industries-query.dto.ts
```diff
@@ -6,13 +6,13 @@ export class GetSectorIndustriesQueryDto {
   @IsString()
   @MaxLength(4)
   @Matches(/^\d{4}$/)
-  @ApiProperty({ example: '1010' })
+  @ApiProperty({ required: false, example: '1010', description: 'Search by UKEF Sector id, returns multiple Industries' })
   public ukefSectorId: string;
 
   @IsOptional()
   @IsString()
   @MaxLength(5)
```
abhi-markan, line 14: Can we also add `@Min(2)`, or a number you think is a common min standard please?

src/modules/sector-industries/dto/get-sector-industries-query.dto.ts
```diff
@@ -6,13 +6,13 @@ export class GetSectorIndustriesQueryDto {
   @IsString()
   @MaxLength(4)
```
abhi-markan, line 7: Can we also add `@Min(2)`, or a number you think is a common min standard please?

src/modules/yield-rates/yield-rates.controller.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    yieldRatesController = app.get<YieldRatesController>(YieldRatesController);
+    yieldRatesService = app.get<YieldRatesService>(YieldRatesService);
+  });
+
+  it('should be defined', () => {
+    expect(yieldRatesController).toBeDefined();
+  });
+
+  describe('find()', () => {
+    it('should call service find function', () => {
```
abhi-markan, line 60: Can we please add more test case with `searchDate` as:
* `ABC`
* `123`
* `!"£!"£`
* `  `
* `0000-00-00`
* `00-00-00`

test/yield-rates/yield-rates.api-test.ts
```diff
@@ -0,0 +1,80 @@
...
+  it(`GET /yield-rates?searchDate=2023-03-02`, async () => {
+    const { status, body } = await api.get('/yield-rates?searchDate=2023-03-02');
+    expect(status).toBe(200);
+    expect(body).toEqual(expect.arrayContaining([expect.objectContaining(yieldRateSchema)]));
+
+    // Field effectiveTo should NOT be set to MAXIMUM LIMIT
+    expect(body).not.toEqual(expect.arrayContaining([expect.objectContaining({ ...yieldRateSchema, effectiveTo: DATE.MAXIMUM_TIMEZONE_LIMIT })]));
+  });
+
+  it(`Expect 404 for GET /yield-rates?searchDate=2010-03-14`, async () => {
```
abhi-markan, line 48: Can we please add a single lint comment why this endpoint will return `404`. Thanks

test/yield-rates/yield-rates.api-test.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(body.message).toContain('searchDate must be a valid ISO 8601 date string');
+  });
+
+  it(`Expect 400 for GET /yield-rates?searchDate=null`, async () => {
+    const { status, body } = await api.get('/yield-rates?searchDate=null');
+    expect(status).toBe(400);
+    expect(body.message).toContain('searchDate must be a valid ISO 8601 date string');
+  });
+
+  it(`Expect 400 for GET /yield-rates?searchDate=undefined`, async () => {
```
abhi-markan, line 71: Can we please add a test case with `searchDate` on outeredges i.e. `9999-12-31`. Thanks

abhi-markan (changes requested):
Good work and nice test coverage, only few comments to address please. Thanks

src/modules/yield-rates/yield-rates.service.ts
avaitonis: Done. Tested and changed for all TypeORM responses. I left Stored Procedure response checks unchanged, maybe SP can return error object, not array.

test/yield-rates/yield-rates.api-test.ts
IgnatG: just to add, can we get the test to read something like
```
it (`returns 404 for ...`) 
```

test/yield-rates/yield-rates.api-test.ts
avaitonis: I added test for 9999-12-31 (404) and 9999-12-30 (last date for active rates)

test/yield-rates/yield-rates.api-test.ts
avaitonis: Added: UKEF at the moment has yield rates since 2010-03-15, maybe some old data will be retired in future.

src/modules/yield-rates/yield-rates.controller.test.ts
avaitonis: Added

src/modules/numbers/dto/get-numbers-query.dto.ts
avaitonis: Added @Min(1). 
We could limit accepted values in future, at the moment passing 3,4,5,6,7 allows testing and using logic in SP.

src/modules/numbers/dto/get-numbers-query.dto.ts
avaitonis: Same as above, Added @min(1).

src/modules/sector-industries/dto/get-sector-industries-query.dto.ts
avaitonis: Changed MaxLength to Length, this tests min and max.

Length validation duplicates regex check, regex is more precise because just digits are allowed.

abhi-markan (dismissed):

test/yield-rates/yield-rates.api-test.ts
avaitonis: Done

abhi-markan (approved):

PR #78, opened 2023-03-29
feat(APIM-162): API management instance

abhi-markan:
## Introduction
An API management gateway is a necessity in order to authenticate, authorize, cache and throttle endpoints. APIM will also provide a façade for the back-end endpoint hosting containers.

## Resolution
* Azure API management instance creation and configuration, APIM `sku` a.k.a plan will be set respective to the targeted deployment environment. `Developer` for development and staging, whereas a pre-defined plan for `production`.
* APIM and CA configuration.
* Enhanced VNET peering creation.
* Naming convention update.
* Base docker image updated to `19.9`.
* Updated `@ApiProperty` from `examples` array to `example` string, due to OpenAPI 3.0 validation error.
* Added `jsonDocumentUrl` for custom direct OpenAPI JSON export.

## Miscellaneous
* Spell fixes
* Deployment script: added staging
* Added GHA files to deployment paths for both `infrastructure` and `deployment`.
* Various dependencies updates.

## Future
* Import APIM policies as IaC.


avaitonis (dismissed):
LGTM

src/modules/constants/entities/constants-spi.entity.ts
```diff
@@ -16,7 +16,7 @@ export class ConstantSpiEntity {
       from: (value: string) => value.toString(),
     },
   })
-  @ApiProperty({ examples: ['A', 'B', 'C', 'Quality of Product', 'Percentage of Cover'] })
+  @ApiProperty({ example: 'A, B, C, Quality of Product, Percentage of Cover' })
```
lmarrai-sw, line 19: We'll have to pick only one of these to be our example if we can't use an array. If we think it's important to list all of these then we could pick one as an `example` and list the others in the `description` instead

abhi-markan: Happy with that approach, @avaitonis what are your thoughts on above pleasE?

.github/workflows/infrastructure.yml
```diff
@@ -23,150 +23,418 @@
...
+            --display-name 'MDM' \
+            --description 'Master database management endpoints' \
+            --service-url https://$(az containerapp show --name ca-${{ env.product }}-mdm-${{ env.target }}-001 --query properties.latestRevisionFqdn -o tsv) \
+            --path 'mdm' \
+            --service-name apim-${{ env.environment }}-${{ env.target }}-001 \
+            --specification-format OpenApi \
+            --api-type http \
+            --protocols https \
+            --subscription-required true \
+            --specification-url https://$(az containerapp show --name ca-${{ env.product }}-mdm-${{ env.target }}-001 --query properties.latestRevisionFqdn -o tsv)/openapi/json
```
lmarrai-sw, line 381: If we're choosing the specification URLs anyway then we don't need to have it at `/openapi/json` anymore, we could keep it at `docs-json`. I think Gabriel only wanted to change it to match one of Azure's expected URLs so that we could import a container app directly

src/swagger.ts
```diff
@@ -39,6 +39,8 @@ export const SwaggerDocs = (app: INestApplication) => {
   );
 
   const options = {
+    // Swagger introduces custom URLs for JSON export https://github.com/nestjs/swagger/pull/2273
```
lmarrai-sw, line 42: I don't think we need the comment here

.github/workflows/development.yml
```diff
@@ -107,8 +109,9 @@ jobs:
       with:
         inlineScript: |
           az containerapp update \
-          --name ca-mdm-${{ env.product }}-${{ env.environment }}-001 \
+          --name ca-${{ env.product }}-mdm-${{ env.environment }}-001 \
           --image ${{ env.acr }}/${{ env.name }}:${{ env.from }} \
+          --revision-suffix v${{ github.run_number }} \
```
lmarrai-sw, line 114: I'd usually use the git commit hash as a suffix - is it going to be easy enough to tell which commit of the code is deployed to the container app?

.github/workflows/infrastructure.yml
```diff
@@ -23,150 +23,418 @@
...
+            --registry-username $(az acr credential show -n cr${{ env.product }}${{ env.target }}001 --query username | tr -d '"') \
+            --cpu 1.0 \
+            --memory 2.0Gi \
+            --min-replicas 1 \
+            --max-replicas 4 \
+            --ingress external \
+            --target-port ${{ vars.MDM_PORT }} \
+            --revisions-mode multiple \
+            --transport auto \
+            --tags ${{ env.tags }}
```
lmarrai-sw, line 220: I assume you're already aware, but just in case you forget, I think you'll want to delete these at some point

.github/workflows/infrastructure.yml
```diff
@@ -23,150 +23,418 @@
...
 # 1. Add delete locks
 # 2. Add log workspace
 # 3. Add Insight
 # 4. Add monitor
-# 5. Add API management service
-# 6. Add Key vault
-# 7. Add VNET peer (x-subscription RBAC SP)
+# 5. Add Key vault
 
 name: Infrastructure 🔨
```
lmarrai-sw, line 32: This is quite a long file, is it possible to split it up at all to make it more readable?

lmarrai-sw (commented):
Thanks for this Abhi! I've left some comments but feel free to merge first and deal with them as a separate PR if that'd be easier

.github/workflows/infrastructure.yml
abhi-markan: I introduced above URL since it does not require any basic authentication also I think it is sets template for any future exports in different format i.e. `/openapi/yaml` and etc.

src/swagger.ts
```diff
@@ -39,6 +39,8 @@ export const SwaggerDocs = (app: INestApplication) => {
...
+    jsonDocumentUrl: 'openapi/json',
```
IgnatG, line 43: Isn't this `openapi.json`?

.github/workflows/infrastructure.yml
```diff
@@ -23,150 +23,418 @@
...
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: Azure defaults ✨
+        uses: Azure/cli@1.0.4
+        with:
+          inlineScript: |
+            # Basic
+            az configure --defaults location=${{ vars.REGION }}
+            az configure --defaults group=rg-${{ env.product }}-${{ env.target }}-001
```
IgnatG, line 96: Can we use `001` from a variable? Something like... `version`. This because we may need to create new variations for it in the futur. I.e. `002`

.github/workflows/infrastructure.yml
IgnatG: Agreed. But it will change in the future. Each Microservice will have its own infrastructure and deployments. So the file will get smaller. 
I think that can take place as part of a different PR

.github/workflows/infrastructure.yml
IgnatG: Can I check why we need to delete them?

.github/workflows/infrastructure.yml
IgnatG: This will probably be for a different PR, but we do need the authentication part 🙂 otherwise anyone can bypass it and execute the codebase 

But we can change it in a separate PR to use an authentication header rather than username/password

abhi-markan: Only APIM will be able to access the above link, since CA has IP restriction in place. I tried sending the headers and unfortunately headers cannot be send as part of `speification-url`.

src/swagger.ts
abhi-markan: It can be any link, I decided to follow modern URL convention instead of hard extensions.
We can all further customise the above if requirement arises for a specific JSON export i.e, `currency/openapi/json`

.github/workflows/infrastructure.yml
abhi-markan: Good shout, will do.

src/modules/constants/entities/constants-spi.entity.ts
abhi-markan: Have agreed with @avaitonis on the pushed change.

src/swagger.ts
abhi-markan: Have removed and added brief description to `README.md` under `Swagger` section.

.github/workflows/infrastructure.yml
abhi-markan: Have introduced `var.VERSION` variable, this can also be further used for iterative infrastructure.

.github/workflows/infrastructure.yml
lmarrai-sw: I assumed we'd want the infrastructure for ESTORE and TFS to be controlled by actions in their own repos, rather than in MDM's repo

.github/workflows/infrastructure.yml
lmarrai-sw: That's fine with me 👍

.github/workflows/development.yml
avaitonis: Its quite common to use truncated commit hash. In Mulesoft deployment action we use
commitHash=$(git rev-parse --short "$GITHUB_SHA")

GITHUB_SHA is populated by github action

.github/workflows/infrastructure.yml
lmarrai-sw: @abhi-markan, the basic authentication for the endpoint is configured in `src/swagger.ts` with the code
```javascript
  app.use(
    ['/docs', '/docs-json', '/docs-yaml'],
    basicAuth({
      challenge: true,
      users: {
        [configService.get<string>('SWAGGER_USER')]: configService.get<string>('SWAGGER_PASSWORD'),
      },
    }),
  );
```
Can we please update the array to not mention `'/docs-json'` anymore since that URL isn't used anymore? (You could also have done that _instead_ of changing the URL for the JSON docs)

@IgnatG when you said
> we do need the authentication part
> otherwise anyone can bypass it and execute the codebase

Users still need to know a valid API key to be able to use the endpoints. Access to the swagger pages (the UI, the JSON file, and the YAML file) only allows the user to _view the documentation_ of the endpoints. I don't know how sensitive we consider this documentation - is IP restriction enough or do we need authentication as well?

abhi-markan: Have removed `/docs-json` and `/docs-yaml` and have also introduced `/openapi/yaml` endpoint.
For endpoints documentation anyone can run it locally to extract API documentations.

.github/workflows/infrastructure.yml
abhi-markan: I have removed respective `TFS` and `ESTORE` IaC commands since they will be handled in their respective repo, they were added purely for testing purpose.

IgnatG (dismissed):
LGTM - great work!

avaitonis (approved):
LGTM

lmarrai-sw (dismissed):
LGTM, thanks @abhi-markan

avaitonis (approved):

avaitonis (approved):

PR #97, opened 2023-04-06
refactor: code refactor. + updated the  documentation

IgnatG:
### Changelog
- updated the types from `string` to `Dates` (where applicable)
- updated the documentation
- updated the NPM packages


src/modules/currencies/entities/currency.entity.ts
```diff
@@ -22,24 +23,24 @@ export class CurrencyEntity {
   isoCode: string;
 
   @Column({ name: 'DATE_CREATED_DATETIME', type: 'timestamp' })
-  @ApiProperty({ example: '2010-03-15 16:13:34.000' })
+  @ApiProperty({ example: '2017-03-06T16:20:17' })
   @Transform(({ value }) => format(value, "yyyy-MM-dd'T'HH:mm:ss"))
```
avaitonis, line 27: Should we remove @Transform here and use new date format "2021-01-26T23:59:59.000Z"

src/modules/sector-industries/entities/sector-industry.entity.ts
```diff
@@ -87,15 +87,15 @@ export class SectorIndustryEntity {
   @ApiProperty({ example: '2010-03-15 00:00:00.000' })
   created: Date;
 
-  @Column({ name: 'DATE_LAST_UPDATED_DATETIME' })
-  @ApiProperty({ example: 0, description: '2010-03-15 00:00:00.000' })
+  @Column({ name: 'DATE_LAST_UPDATED_DATETIME', type: 'timestamp' })
+  @ApiProperty({ example: '2010-03-15 00:00:00.000' })
```
avaitonis, line 91: Result is in format: 2017-04-01T00:00:00.000Z
Example could be in same format.

IgnatG: good shout - updated

src/modules/currencies/entities/currency.entity.ts
IgnatG: good shout - updated

PR #147, opened 2023-05-01
fix(APIM-194): fixed the `oecdRiskCategory` differences  in `/markets` endpoint

IgnatG:
#### Changelog
- fixed the differences in `oecdRiskCategory` response
- updated npm packages

### DIFF
--- a/src/modules/markets/markets.controller.ts
+++ b/src/modules/markets/markets.controller.ts
@@ -19,12 +19,7 @@ export class MarketsController {
     description: 'All markets',
     type: MarketEntity,
   })
-  async findAll(@Query() query: MarketsQueryDto): Promise<MarketEntity[]> {
-    const results = await this.marketService.findAll(query.active);
-    const mappedResults = results.map((market: any) => ({
-      ...market,
-      oecdRiskCategory: parseInt(market.oecdRiskCategory.replace(/\D/g, ''), 10),
-    }));
-    return mappedResults;
+  findAll(@Query() query: MarketsQueryDto): Promise<MarketEntity[]> {
+    return this.marketService.findAll(query.active);
   }
 }
--- a/src/modules/markets/markets.service.ts
+++ b/src/modules/markets/markets.service.ts
@@ -17,21 +17,25 @@ export class MarketsService {
 
   async findAll(active?: string): Promise<MarketEntity[]> {
     try {
-      let spResults = await this.marketsRepository.query('CIS_USP_READ_MARKETS');
+      let results = await this.marketsRepository.query('CIS_USP_READ_MARKETS');
 
       if (active) {
         if (active === 'Y') {
-          spResults = spResults.filter((results: { ACTIVE_IND: string }) => results.ACTIVE_IND === 'Y');
+          results = results.filter((results: { ACTIVE_IND: string }) => results.ACTIVE_IND === 'Y');
         } else {
-          spResults = spResults.filter((results: { ACTIVE_IND: string }) => results.ACTIVE_IND !== 'Y');
+          results = results.filter((results: { ACTIVE_IND: string }) => results.ACTIVE_IND !== 'Y');
         }
       }
 
       const fieldMap = DbResponseHelper.getApiNameToDbNameMap(this.marketsRepository);
+      const renamedFields = DbResponseHelper.renameDbResultFields(this.marketsRepository, fieldMap, results);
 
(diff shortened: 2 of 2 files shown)


abhi-markan (dismissed):

src/modules/markets/markets.controller.ts
```diff
@@ -17,9 +17,14 @@ export class MarketsController {
   @ApiResponse({
     status: 200,
     description: 'All markets',
-    type: MarketEntity,
+    type: GetMarketsDto,
```
avaitonis, line 20: This breaks OpenApi spec. Maybe it can keep using MarketEntity, as oecdRiskCategory is number in MarketEntity.

Not sure if we need GetMarketsDto.

lmarrai-sw: I think we need a way to differentiate between the result from the stored procedure (which is a string) and the result we want to give from the API (which is a number). Using 2 different classes is probably the easiest way to do it

Either `MarketEntity` or `MarketsService` needs a change though because `marketsService.findAll` is supposed to return `Promise<MarketEntity[]>` but it returns `oecdRiskCategory` as a string when `MarketEntity` says it should be a number

src/modules/markets/entities/market.entity.ts
```diff
@@ -4,7 +4,7 @@ import { DATE } from '@ukef/constants';
 import { Column, Entity, PrimaryColumn } from 'typeorm';
 
 @Entity({
-  name: 'ODS_MARKETS',
+  name: 'CIS_USP_READ_MARKETS',
   schema: 'dbo',
```
lmarrai-sw, line 8: I'm not sure how we've been using `Entity` `name` and `schema` in MDM up until now, but the TypeORM docs say that the `name` is supposed to be the table name.

(Since this class is for the result of a stored procedure rather than for a table, I don't think it makes sense to have an `Entity` decorator at all but I realise that decision would affect a lot of MDM so we probably don't want to change it at this late stage)

src/modules/markets/dto/get-markets.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+  public marketRiskAppetitePublicDesc: string;
+  public geographicalRegionId: number;
+  public geographicalRegionDesc: string;
+  public sovereignRiskProvision: number;
+  public ESRAClassificationId: number;
+  public ESRAClassificationDesc: string;
+  public shortTermCoverAvailabilityId: number;
+  public shortTermCoverAvailabilityDesc: string;
+  public NBIIssue: string;
+  public active: string;
```
lmarrai-sw: We don't need the `public` keyword

src/modules/markets/entities/market.entity.ts
lmarrai-sw: @IgnatG to be clear, I'm happy for you to do nothing about this comment - feel free to mark it as resolved if that's your plan

src/modules/markets/markets.controller.ts
lmarrai-sw: @IgnatG sorry, I didn't see your update before my comment 

My concern with using only 1 class for it is that it means our type system is wrong and incomplete:
- the return type of `MarketsService` is incorrect because it returns the wrong type of `oecdRiskCategory`
- we don't have a class / type to describe the result from the stored procedure

avaitonis: "the return type of MarketsService is incorrect because it returns the wrong type of oecdRiskCategory"
  * Maybe moving type change to MarketsService would fix issue above
  
"we don't have a class / type to describe the result from the stored procedure" 
  * It is the case for all SPs. We don't have class/type to describe SP result, but TypeOrm entity is used describe both, also to rename all SP field names to API field names. This is not standard approach.

src/modules/markets/entities/market.entity.ts
IgnatG: yes, that would be my expectation as well - that the name is the table name. the only reason why it changed is to match the name of the stored procedure. The long term plan will be to replace the Stored Procedure with the actual tables. I'll create a ticket and add it to Jira

src/modules/markets/markets.controller.ts
lmarrai-sw: Thanks @IgnatG, I'm happy with the change. I'll leave it to @avaitonis to resolve this comment when he is happy too

lmarrai-sw (approved):

PR #152, opened 2023-05-03
fix(deployment): generic deployment script

abhi-markan:
## Introduction
A generic deployment script

## Resolution
* Refer to the `TARGET` and `VERSION` from GH environment variables.
* Simplified YML scripts.


IgnatG (approved):
LGTM

PR #159, opened 2023-05-07
fix: lint fixes based on sonarcloud

IgnatG:
### Changelog
- fixed: Copying recursively might inadvertently add sensitive data to the container.
- fixed: Omitting --ignore-scripts can lead to the execution of shell scripts
- fixed: Promises must be awaited


Dockerfile
```diff
@@ -11,20 +11,20 @@ RUN rm -rf /var/cache/apk/*
...
+COPY --chown=node:node tsconfig*.json .
 
 # Build with all dependencies
 RUN npm run build
 
 # Lean NPM - Only install `dependencies`
 # `devDependencies` will still be resolved inside `package-lock.json`,
 # however they will not be installed inside `node_modules` directory.
-RUN npm ci --legacy-peer-deps --omit=dev
+RUN npm ci --legacy-peer-deps --omit=dev --ignore-scripts
```
abhi-markan, line 27: I have tried this in the past and it has caused `npm run build` to fail, were you able to build the container without any errors locally?

Dockerfile
```diff
@@ -11,20 +11,20 @@ RUN rm -rf /var/cache/apk/*
...
-COPY --chown=node:node package.json .
-COPY --chown=node:node package-lock.json .
-RUN npm ci --legacy-peer-deps
+COPY --chown=node:node package*.json .
+RUN npm ci --legacy-peer-deps --ignore-scripts
 RUN npm cache clean --force
 
-COPY --chown=node:node . .
+COPY --chown=node:node src src
+COPY --chown=node:node tsconfig*.json .
```
abhi-markan, line 19: This certainly makes code clean, but should we not be explicit with exact file names?

Dockerfile
```diff
@@ -11,20 +11,20 @@ RUN rm -rf /var/cache/apk/*
...
 # NPM
-COPY --chown=node:node package.json .
-COPY --chown=node:node package-lock.json .
-RUN npm ci --legacy-peer-deps
+COPY --chown=node:node package*.json .
+RUN npm ci --legacy-peer-deps --ignore-scripts
 RUN npm cache clean --force
 
-COPY --chown=node:node . .
+COPY --chown=node:node src src
```
abhi-markan, line 18: We also need to copy below build dependant file please:

* `nest-cli.json`

Dockerfile
IgnatG: I've tried it locally and it's working for me. I wonder if it's WSL otherwise?

Dockerfile
IgnatG: in theory yes. the `*` is just a wild card to copy every file that starts with tsconfig - this is to reduce repetition. I don't mind either way tbh. let me know which one works best for him; happy to update

Dockerfile
IgnatG: updated now

Dockerfile
IgnatG: never mind. it crashed now...
I've removed it

Dockerfile
abhi-markan: Explicit file names please, thanks

IgnatG: updated now

abhi-markan (approved):

PR #167, opened 2023-05-11
feat(APIM-310): added new query parameter 'search' to GET /markets

avaitonis:
### Introduction
- During MDM integration with DTFS we noticed that they use GET /Countries endpoint.
- This PR adds enough functionality to GET /markets endpoint that it can replace GET /countries for DTFS

### Resolution
- Added query parameter 'search' to GET /markets
- Works in similar manner like in Mulesoft implementation
  - parameter is optional
  - search allows partial matches
  - it is not case sensitive
- Node JS keeps getting all markets from SP and filters results later, overhead is minimal as we have just 236 markets
- old endpoint `/countries` was getting data from CEDAR, `/markets` is getting data from CIS DB. They are very similar, but some differences might occur.

### Miscellaneous 
- fixed GET /exposure-period

### DIFF
--- a/src/modules/exposure-period/exposure-period.service.ts
+++ b/src/modules/exposure-period/exposure-period.service.ts
@@ -19,7 +19,7 @@ export class ExposurePeriodService {
       // TODO: SP USP_MDM_READ_EXPOSURE_PERIOD is not using data/tables from DB. Calculation could be moved to Javascript.
       const results = await this.mdmDataSource.query('USP_MDM_READ_EXPOSURE_PERIOD @0, @1, @2', [startDate, endDate, productGroup]);
 
-      if (!results.Length || typeof results[0].EXPOSURE_PERIOD === 'undefined') {
+      if (!results.length || typeof results[0].EXPOSURE_PERIOD === 'undefined') {
         throw new InternalServerErrorException('No exposure period result from USP_MDM_READ_EXPOSURE_PERIOD');
       }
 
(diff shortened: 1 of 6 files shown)


IgnatG (approved):
LGTM

src/modules/exposure-period/exposure-period.service.ts
```diff
@@ -19,7 +19,7 @@ export class ExposurePeriodService {
       // TODO: SP USP_MDM_READ_EXPOSURE_PERIOD is not using data/tables from DB. Calculation could be moved to Javascript.
       const results = await this.mdmDataSource.query('USP_MDM_READ_EXPOSURE_PERIOD @0, @1, @2', [startDate, endDate, productGroup]);
 
-      if (!results.Length || typeof results[0].EXPOSURE_PERIOD === 'undefined') {
+      if (!results.length || typeof results[0].EXPOSURE_PERIOD === 'undefined') {
```
abhi-markan, line 22: Do we need index `0` void check before accessing it?

abhi-markan (changes requested):
Good work and good test coverage.
Just one minor comment to address please.

src/modules/exposure-period/exposure-period.service.ts
IgnatG: `results` will always be an array - so the length will either be 0 or more

avaitonis: Marking as resolve, we used to use index 0, but don't do it anymore.

PR #180, opened 2023-05-15
chore: only lint once on commit

lmarrai-sw:
## Introduction
The on-commit hook takes a bit of time to run, especially for members on the team on AVDs, so I looked into if we could make it quicker. I noticed that we were linting twice: we were using `lint-staged` to `eslint --fix` all the staged files and then running `npm run lint` on the whole project after.

## Resolution
I removed npm run lint from the commit hook so we only lint once (and only the staged files) in the on-commit hook


IgnatG (approved):
LGTM

PR #210, opened 2023-05-26
feat(APIM-261): API import post deployment

abhi-markan:
## Introduction
Post deployment, APIM API service URL and latest endpoints are not imported as part of CICD.

## Resolution
* APIM API import command updates API service URL (latest revision FQDN).
* Same command also imports the API operations using OpenAPI 3 specification `/openapi/json` from the latest CA revision FQDN.
* Conditional local VNET peering to `pre-prod` or `prod` VNET.

## Miscellaneous
* Various outdated dependencies have been updated.

### DIFF
--- a/src/modules/healthcheck/healthcheck.controller.ts
+++ b/src/modules/healthcheck/healthcheck.controller.ts
@@ -22,7 +22,7 @@ export class HealthcheckController {
     private readonly mem: MemoryHealthIndicator,
   ) {}
 
-  @Get('ready')
+  @Get('API micro-service health check performs dependent databases connection and memory heap check')
   @HealthCheck()
   @ApiResponse({ status: 200, description: 'Check if this api can access dependencies' })
   @ApiOperation({ summary: 'ready' })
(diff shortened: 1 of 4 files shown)


src/modules/healthcheck/healthcheck.controller.ts
```diff
@@ -22,7 +22,7 @@ export class HealthcheckController {
     private readonly mem: MemoryHealthIndicator,
   ) {}
 
-  @Get('ready')
+  @Get('API micro-service health check performs dependent databases connection and memory heap check')
```
IgnatG, line 25: The `Get` decorator is used to create a path i.e. /api/v1/ready
This from above looks like the description of it

abhi-markan: Ah apologies I'll double check I was planning to update the description

abhi-markan: Have rectified.

IgnatG (approved):
LGTM

PR #313, opened 2023-07-12
fix(APIM-531): fix swagger regex field

AlexBramhill:


lmarrai-sw (approved):
LGTM thanks @AlexBramhill 
(there's a linter issue which we're already aware of and will be fixed in a separate PR)

PR #324, opened 2023-07-17
feat(APIM-344): handle sensitive details in log messages

avaitonis:
### Introduction
Add functionality to remove sensitive data before writing it to the logs.

### Resolution
  - Redact errors in PinoLogger logs by unsetting some attributes in error. 
    - Functionality copied from TFS-API. PinoLogger config gets new attribute `redact`.
  - Bootstrap might have errors with DB domain. Bootstrap errors are solved by using custom logger in src/main.ts
    - Functionality is done using custom logger ConsoleLoggerWithRedact
  - Redact errors in TypeORM error message strings.
    - Error strings are modified using PinoLogger hook logMethod. Hook is set in main.module.ts.
    - This is done using helper function redactStringsInLogArgs.

### Miscellaneous 
- Changed how Logger is loaded in the services. Now they use dependency injection.
- New environment variable `REDACT_LOGS`

### Todo
  - Try catch could be removed from services, but default global exception handler logs smaller error dump.


src/logging/console-logger-with-redact.ts
```diff
@@ -0,0 +1,20 @@
...
+  error(message: any, stack?: string, context?: string) {
+    let cleanStack = stack;
+    if (typeof stack == 'string') {
+      this.stringPatternsToRedact.forEach((redact) => {
+        cleanStack = stack.replace(redact.searchValue, redact.replaceValue);
+      });
+    }
+    super.error(message, cleanStack, context);
+  }
+}
```
lmarrai-sw, line 20: It would be nice to have tests for this :)

src/main.ts
```diff
@@ -1,14 +1,15 @@
...
+import { REDACT_STRINGS } from './constants';
+import { ConsoleLoggerWithRedact } from './logging/console-logger-with-redact';
 
 const main = async () => {
-  const nestApp: NestApplication = await NestFactory.create(MainModule, { bufferLogs: true });
+  const logger = new ConsoleLoggerWithRedact(REDACT_STRINGS);
+  const nestApp: NestApplication = await NestFactory.create(MainModule, { logger, bufferLogs: false });
   const app = new App(nestApp);
 
-  const logger = new NestLogger();
```
lmarrai-sw, line 11: Do we lose out on much functionality by replacing the `NestLogger` with the `ConsoleLoggerWithRedact`?

src/modules/interest-rates/interest-rates.service.ts
```diff
@@ -1,23 +1,27 @@
...
+    private readonly logger: PinoLogger,
+    private readonly config: ConfigService,
   ) {}
 
   findAll(): Promise<InterestRatesEntity[]> {
     try {
       return this.interestRates.find({ where: { effectiveTo: Equal(new Date(DATE.MAXIMUM_LIMIT)) } });
     } catch (err) {
-      this.logger.error(err);
+      this.logger.error(redactError(this.config.get<boolean>('app.redactLogs'), REDACT_STRING_PATHS, REDACT_STRINGS, err));
```
lmarrai-sw, line 24: Since you're using dependency injection for the logger now, we should be able to make our own logger that:
- has a PinoLogger itself (we can dependency inject it in)
- has a ConfigService (or just the `app` config) itself (we can dependency inject it in)
- has `error` and `warn` methods that take an error, call the `redactError` function on it, and then the PinoLogger's error/warn method with the result

so that we use this `redactError` function in just 1 place instead of in every service (and so we don't risk someone forgetting to use it in future)
What do you think?

src/logging/log-keys-to-redact.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    },
+  };
+
+  describe('when redactLogs is false', () => {
+    it('returns an empty array', () => {
+      expect(logKeysToRedact({ redactLogs: false, ...options })).toStrictEqual([]);
+    });
+  });
+
+  describe('when redactLogs is true', () => {
```
lmarrai-sw, line 33: I think there are more tests we can test here, like testing that the `dbError` keys are included in the keys to redact

src/helpers/redact-errors.helper.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (value) {
+      const safeValue = redactString(redactStrings, value);
+      set(err, path, safeValue);
+    }
+  });
+  return err;
+};
+
+const redactString = (redactStrings: { searchValue: string | RegExp; replaceValue: string }[], string: string): string => {
+  let safeSTring: string = string;
```
lmarrai-sw, line 24: small typo here in `safeSTring`

src/logging/console-logger-with-redact.ts
avaitonis: Was planning to add them and now it is done.

src/modules/interest-rates/interest-rates.service.ts
avaitonis: We don't use PinoLogger directly, we use nestjs module nestjs-pino as a convenient way to use PinoLogger. 
Wasn't sure what duplicating nestjs-pino was good idea, and found that PinoLogger has `hooks` that allows us redact all log messages.

Hook is defined in main.module.ts.
```
hooks: {
            logMethod(inputArgs: any, method) {
              return method.apply(this, redactStringsInLogArgs(config.get<boolean>('app.redactLogs'), REDACT_STRING_PATHS, REDACT_STRINGS, inputArgs));
            },
          },
```

src/logging/log-keys-to-redact.test.ts
avaitonis: I added dbError key tests

src/helpers/redact-errors.helper.ts
avaitonis: Thanks, changed

.env.sample
```diff
@@ -2,6 +2,7 @@
 PORT=
 
 LOG_LEVEL=debug
+REDACT_LOGS=false
```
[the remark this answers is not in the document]
avaitonis: Created task - https://ukef-dtfs.atlassian.net/browse/APIM-566

src/main.ts
```diff
@@ -6,8 +7,10 @@ import { REDACT_STRINGS } from './constants';
 import { ConsoleLoggerWithRedact } from './logging/console-logger-with-redact';
 
 const main = async () => {
-  const logger = new ConsoleLoggerWithRedact(REDACT_STRINGS);
-  const nestApp: NestApplication = await NestFactory.create(MainModule, { logger, bufferLogs: false });
+  // If REDACT_LOGS is true use ConsoleLoggerWithRedact. ConsoleLoggerWithRedact is used just if `NestFactory.create` is fails completely.
```
lmarrai-sw, line 10: I don't think we need the last `is` in this sentence :)

src/main.ts
```diff
@@ -6,8 +7,10 @@ import { REDACT_STRINGS } from './constants';
...
+  // If `NestFactory.create` doesn't fail completely, then buffered logs are passed to PinoLogger. NestLogger and ConsoleLoggerWithRedact are not used.
+  const logger = process.env.REDACT_LOGS !== 'false' ? new ConsoleLoggerWithRedact(REDACT_STRINGS) : new NestLogger();
+  const nestApp: NestApplication = await NestFactory.create(MainModule, { logger, bufferLogs: true });
```
lmarrai-sw, line 13: Why did we change `bufferLogs` to true?

src/modules/interest-rates/interest-rates.service.ts
lmarrai-sw: I like the `hooks` idea, great find @avaitonis!

src/helpers/redact-strings-in-log-args.helper.ts
```diff
@@ -1,29 +1,31 @@
 import { get, set } from 'lodash';
 
 // This helper function is used to redact sensitive data in Error object strings.
-export const redactError = (
+export const redactStringsInLogArgs = (
   redactLogs: boolean,
   redactPaths: string[],
   redactStrings: { searchValue: string | RegExp; replaceValue: string }[],
-  err: any,
+  args: any,
```
lmarrai-sw, line 8: Can we put any stricter type that `any` here? Do we know it's an array at least?

lmarrai-sw (commented):
Thanks @avaitonis

src/main.ts
avaitonis: `Is` definitely is not required

src/helpers/redact-strings-in-log-args.helper.ts
avaitonis: I looked at pico source code and documentation, looks like type is universal, only limit is that it is an array. Changed to any[]

src/main.ts
```diff
@@ -3,12 +3,16 @@ import { NestApplication, NestFactory } from '@nestjs/core';
...
 
 import { App } from './app';
+import { REDACT_STRINGS } from './constants';
+import { ConsoleLoggerWithRedact } from './logging/console-logger-with-redact';
 
 const main = async () => {
-  const nestApp: NestApplication = await NestFactory.create(MainModule, { bufferLogs: true });
+  // If REDACT_LOGS is true use ConsoleLoggerWithRedact. ConsoleLoggerWithRedact is used just if `NestFactory.create` fails completely.
+  // If `NestFactory.create` doesn't fail completely, then buffered logs are passed to PinoLogger. NestLogger and ConsoleLoggerWithRedact are not used.
+  const logger = process.env.REDACT_LOGS !== 'false' ? new ConsoleLoggerWithRedact(REDACT_STRINGS) : new NestLogger();
```
IgnatG, line 12: isn't this `boolean`? rather than `string`?

avaitonis: All variables in process.env are strings. We need to parse them to boolean and numbers.

Usually we get configuration from Nestjs configService, but it is not available yet in this place, configuration service would return correct type `boolean`.

PR #365, opened 2023-07-28
feat(APIM-538): update single line log format env var

AlexBramhill:
# Introduction
fixed typo causing the env var for single line log format to not work


lmarrai-sw (approved):
Good spot, thanks @AlexBramhill

abhi-markan (approved):

PR #967, opened 2024-07-30
fix(company-house): company house API response

abhi-markan:
## Introduction :pencil2:
Company house no longer returns expected body (below) for a non-existent company with an 8 digit CRN.

```json
{
    "errors": [
        {
            "type": "ch:service",
            "error": "company-profile-not-found"
        }
    ]
}
```

* Updated non-existent company house to CRN to a 7 digit, which still returns the expected body for a `404` response.
* Refactored `createWrapCompaniesHouseHttpGetErrorCallback`.
* Introduced two additional help functions.
* Updated test cases.

## Miscellaneous ➕ 
* Dependencies updates
* Updated snapshot.


AlexBramhill (approved):

AlexBramhill (approved):

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,21 +9,25 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
 export const createWrapCompaniesHouseHttpGetErrorCallback =
   ({ messageForUnknownError, knownErrors }: { messageForUnknownError: string; knownErrors: KnownErrors }): CompaniesHouseHttpErrorCallback =>
   (error: Error) => {
-    if (error instanceof AxiosError && error.response && typeof error.response.data === 'object') {
-      const errorResponseData = error.response.data;
-      let errorMessage: string;
+    if (error instanceof AxiosError) {
```
AlexBramhill, line 12: could we guard clause this please to remove some complexity (`if (!isAxiosError(error)... bla` -- Also by doing `isAxioxError` it helps with typescript inferring the type better -- should make it complain less about accessing `error.response` etc

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,21 +9,25 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
-
-        knownErrors.forEach(({ caseInsensitiveSubstringToFind, throwError }) => {
-          if (errorMessageInLowerCase.includes(caseInsensitiveSubstringToFind.toLowerCase())) {
+      /**
+       * CH API will throw empty response body with `404`
+       * upon sending a valid non-existent 8 digit CRN.
+       *
+       * Otherwise `404` with an expected response body.
+       */
+      if (message || status) {
```
AlexBramhill, line 28: I think we should/could remove this check as it's coupling the inner workings of our error handling instances with this function. In an ideal world, this function does not care about the errors it's handling -- all it needs to know about the `knownErrors` is that it can call `checkHasError` on it.

The current implimentation does save some time technically, but I think at the cost of making the code more coupled where it doesn't need to be, and less extendable later... What do you think?

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,21 +9,25 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
+        knownErrors.forEach(({ checkHasError, throwError }) => {
+          if (checkHasError(error)) {
```
AlexBramhill, line 30: Could this be simplified into `checkHasError(error) && throwError(error)`

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,21 +9,25 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
 export const createWrapCompaniesHouseHttpGetErrorCallback =
   ({ messageForUnknownError, knownErrors }: { messageForUnknownError: string; knownErrors: KnownErrors }): CompaniesHouseHttpErrorCallback =>
   (error: Error) => {
-    if (error instanceof AxiosError && error.response && typeof error.response.data === 'object') {
-      const errorResponseData = error.response.data;
-      let errorMessage: string;
+    if (error instanceof AxiosError) {
+      let message: string;
```
AlexBramhill, line 13: `string | undefined` as we've got an `if else` on lines `16-19` that might not fire if neither of the conditions are true

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,21 +9,25 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
+      const status: number = error?.response?.status;
```
AlexBramhill, line 14: `const status: number | null= error?.response?.status;`.... oddly 

https://github.com/axios/axios/blob/8966ee7ea62ecbd6cfb39a905939bcdab5cf6388/lib/core/AxiosError.js#L50

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
+import { HttpStatus } from '@nestjs/common';
+import { isAxiosError } from 'axios';
+
+/**
+ * Checks if the error object is an AxiosError and has a response status matching the specified status.
+ * @param {Object} options - An object containing the error and status.
```
AlexBramhill, line 6: We shouldn't comment types here as it overrides and mucks around with typescript -- It's good to list the params name with a description if it helps tho :)

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Error} options.error - The error object to check.
+ * @param {HttpStatus} options.status - The status to compare against.
+ * @returns {boolean} - Returns true if the error is an AxiosError and has a matching response status, otherwise false.
+ */
+export const statusCheck = ({ error, status }: { error: Error; status: HttpStatus }): boolean => isAxiosError(error) && error.response?.status === status;
```
AlexBramhill, line 11: We know that this error is `AxiosError` on entering here (Although, I do agree with how it's written here and think we should check for axios error in each of these functions,... If that's the case, I made a mistake with this code and the error should be `unknown` until we check for axios error -- as technically that's the type when handling errors)

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Validates the error response against the specified search string.
+ * @param {Object} options - An object containing the error and search string.
+ * @param {Error} options.error - The error object to validate.
+ * @param {string} options.search - The search string to compare against the error message.
+ * @returns {boolean} - Returns true if the error response contains the search string, otherwise false.
+ */
+export const messageCheck = ({ error, search }: { error: Error; search: string }): boolean => {
```
AlexBramhill, line 20: same comments re error -- thats my bad

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
AlexBramhill: This would also allow us to deduplicate the code on lines 16-19 of this file, as it would no longer be needed as we'd be decoupling this function from caring about it.

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!isAxiosError(error) || !error?.response?.data) {
```
AlexBramhill, line 21: Im not sure we need the second boolean check as we check it in the if statements below anyway?

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
...
+    return false;
+  }
+
+  let message: string;
+
+  if (typeof error?.response?.data?.error === 'string') {
+    message = error.response.data.error;
+  } else if (error?.response?.data?.errors[0] && typeof error?.response?.data?.errors[0]?.error === 'string') {
```
AlexBramhill, line 29: does this work
```suggestion
  } else if (typeof error?.response?.data?.errors?.[0]?.error === 'string') {
```

src/helpers/response-status.helper.ts
```diff
@@ -0,0 +1,36 @@
...
+    message = error.response.data.errors[0].error;
+  }
+
+  if (message) {
+    return message.toLowerCase().includes(search.toLowerCase());
+  }
```
AlexBramhill, line 35: need to return a non-happy path otherwise it'll return `void | boolean` 
```suggestion
  return message ? message.toLowerCase().includes(search.toLowerCase()) : false
```

test/companies/get-company-by-registration-number.api-test.ts
```diff
@@ -97,16 +97,13 @@ describe('GET /companies?registrationNumber=', () => {
...
     requestToGetCompanyByRegistrationNumber(companiesHousePath).reply(404, getCompanyCompaniesHouseNotFoundResponse);
 
     const { status, body } = await api.get(mdmPath);
 
     expect(status).toBe(404);
-    expect(body).toStrictEqual({
-      statusCode: 404,
-      message: 'Not found',
-    });
+    expect(body).toStrictEqual({});
```
AlexBramhill, line 106: I would not expect this to change as the error we're throwing hasn't changed?

AlexBramhill (changes requested):

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Have added `&& error?.response` as `404` now returns empty `error.response.data` object.

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Good shout, have refactored.

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Good spot, have added.

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Have removed as this might not be necessary anymore.

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Have removed as this might not be necessary anymore.

src/helpers/response-status.helper.ts
abhi-markan: The error being sent to these functions is of type `Error` under type `KnownError`.

src/helpers/response-status.helper.ts
abhi-markan: This is useful to save any execution if the `response.data` does not exists.

src/helpers/response-status.helper.ts
abhi-markan: Good spot.

src/helpers/response-status.helper.ts
abhi-markan: Good spot, have actually updated to Nullish coalescing operator.

src/helpers/response-status.helper.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  it('should return true when isAxiosError is true and status matches', () => {
+    const error = {
+      response: {
+        status: HttpStatus.BAD_REQUEST,
+      },
+      name: 'Bad request',
+      message: 'Bad request',
+    } as Error;
```
avaitonis, line 22: Should type be AxiosError from 'axios'?

src/helpers/response-status.helper.ts
avaitonis: Looks like recommendations exist to not use jsDoc types in Typescript.

Google style guide - https://google.github.io/styleguide/tsguide.html#jsdoc-type-annotations
```JSDoc type annotations are redundant in TypeScript source code. Do not declare types in @param or @return blocks, do not write @implements, @enum, @private, @override etc. on code that uses the implements, enum, private, override etc. keywords.```

So you could remove {Object}, {Error}, {HttpStatus}, {boolean}

File: src/helpers/response-status.helper.ts (hunk 7 above)
Comment (AlexBramhill) [reply src/helpers/response-status.helper.ts:11]: Odd -- Okay happy to leave this as is

File: src/helpers/response-status.helper.ts (hunk 9 above)
Comment (AlexBramhill) [reply src/helpers/response-status.helper.ts:21]: Cool -- Happy :)

File: src/helpers/response-status.helper.ts [hunk 14]
```diff
@@ -0,0 +1,34 @@
...
+ * @param {string} options.search - The search string to compare against the error message.
+ * @returns {boolean} - Returns true if the error response contains the search string, otherwise false.
+ */
+export const messageCheck = ({ error, search }: { error: Error; search: string }): boolean => {
+  if (!isAxiosError(error) || !error?.response?.data) {
+    return false;
+  }
+
+  let message: string;
+
```
AlexBramhill, line 26: this should be string | undefined :)

src/helpers/response-status.helper.test.ts
AlexBramhill: It should -- This should not be working without this.

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,25 +9,12 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
-      if (errorMessage) {
-        const errorMessageInLowerCase = errorMessage.toLowerCase();
-
-        knownErrors.forEach(({ caseInsensitiveSubstringToFind, throwError }) => {
-          if (errorMessageInLowerCase.includes(caseInsensitiveSubstringToFind.toLowerCase())) {
-            return throwError(error);
-          }
-        });
-      }
+    if (error instanceof AxiosError && error?.response) {
```
AlexBramhill, line 12: I really don't think we need this but we can keep it for now :)

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,25 +9,12 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
+      knownErrors.forEach(({ checkHasError, throwError }) => {
+        if (checkHasError(error) && throwError(error)) {
```
AlexBramhill, line 14: we dont need to check throw error here

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
```diff
@@ -9,25 +9,12 @@ type CompaniesHouseHttpErrorCallback = (error: Error) => ObservableInput<never>;
...
+          return throwError(error);
+        }
```
AlexBramhill, line 16:
```suggestion
        checkHasError(error) && throwError(error)
```

src/helper-modules/companies-house/wrap-companies-house-http-error-callback.ts
abhi-markan: Discussed, keeping it as it is for consistency.

src/helpers/response-status.helper.ts
abhi-markan: Have removed.

avaitonis (approved):
LGTM

PR #1056, opened 2024-11-08
feat(FN-3163): MDM endpoints for automatically creating new DTFS customers in Salesforce (Dun and Bradstreet service only)

natdeanlewissoftwire:
## Introduction :pencil2:
As part of the CCM project, if a customer does exist in SF, we are automatically creating it rather than relying on the manual process currently undertaken by Business Support Group. DTFS will send requests to a new customer/salesforce endpoint for MDM to create a customer in Salesforce. Technical documentation of this process is [here](https://ukef.sharepoint.com/:w:/r/sites/pmo-PRO211/Shared%20Documents/1.%20TO%20USE%20New%20file%20plan/2.%20Tranche%202/03.%20Customer%20and%20Credit%20Migration/Documents/DTFS%20%26%20Salesforce%20Integration/DTFS%20-%20SF%20Integration%20Technical%20Details.docx?d=w75ec8ca455ec492db2a912cc20cea7e7&csf=1&web=1&e=rhuzKr)

The corresponding feature-flagged DTFS pull request is [here](https://github.com/UK-Export-Finance/dtfs2/pull/4021)

This PR adds the Dun and Bradstreet service, and an endpoint to allow us to GET a company from Dun and Bradstreet by their registration number (in our case Companies House Number). This will then be used by [this PR](https://github.com/UK-Export-Finance/mdm-api/pull/1057) as part of the customer creation process.

To be merged before [this PR](https://github.com/UK-Export-Finance/mdm-api/pull/1057)


src/main.module.ts
```diff
@@ -42,10 +42,16 @@ import { LoggingInterceptor } from './logging/logging-interceptor.helper';
             clientRequest: {
               logKey: 'req',
               headersLogKey: 'headers',
+              bodyLogKey: 'data',
```
avaitonis, line 45: Should we use constant BODY_LOG_KEY here?

src/main.module.ts
```diff
@@ -42,10 +42,16 @@ import { LoggingInterceptor } from './logging/logging-interceptor.helper';
...
             },
             outgoingRequest: {
               logKey: OUTGOING_REQUEST_LOG_KEY,
               headersLogKey: HEADERS_LOG_KEY,
+              bodyLogKey: BODY_LOG_KEY,
+            },
+            incomingResponse: {
+              logKey: INCOMING_RESPONSE_LOG_KEY,
+              bodyLogKey: BODY_LOG_KEY,
```
avaitonis, line 54: Is this removing full incoming response body? or just one key in the JSON?
Body still might be nice to log for debugging.

Is this removing full outgoing response body? or just one key in the JSON?
Because Post/body is new functionality for http client it might be fine to remove full body from the log.

src/main.module.ts
natdeanlewissoftwire: Yep, good spot :)

src/main.module.ts
natdeanlewissoftwire: Good point - I've specified just the access token rather than the whole body to redact in the incoming response, and removed the outgoing request change as it's actually not necessary anymore, was from a previous implementation.

src/helper-modules/dun-and-bradstreet/dun-and-bradstreet.service.ts
```diff
@@ -0,0 +1,56 @@
...
+import { DUN_AND_BRADSTREET } from '@ukef/constants';
+import { HttpClient } from '@ukef/modules/http/http.client';
+
+import { DunAndBradstreetConfig } from '@ukef/config/dun-and-bradstreet.config';
+import { createWrapDunAndBradstreetHttpGetErrorCallback } from './wrap-dun-and-bradstreet-http-error-callback';
+
+@Injectable()
+export class DunAndBradstreetService {
+  private readonly httpClient: HttpClient;
+  private readonly encoded_key: string;
```
avaitonis, line 13: Should we use camelCase for all variables? encodedKey, accessToken

natdeanlewissoftwire: Good spot, thanks I'll do that.

PR #1082, opened 2024-12-06
fix(APIM-608): make GovNotify properties optional

abhi-markan:
## Introduction :pencil2:
Make `unsubscribe_link` and `scheduled_for` properties are set to optional for strict schema validation.

* Set the `required` to false.

* Dependencies updates.
* Added additional validation scripts to `package.json`


src/helper-modules/govuk-notify/dto/post-emails-response.dto.ts
```diff
@@ -24,6 +24,7 @@ export class PostEmailsResponseDataContent {
     example: null,
     description: 'Unsubscribe links are not used, so this field should always be null',
     nullable: true,
+    required: false,
```
ttbarnes, line 27: It's strange there are no tests that require updating. Not a blocker.

ttbarnes (approved):

ttbarnes (dismissed):

src/helper-modules/govuk-notify/dto/post-emails-response.dto.ts
abhi-markan: Agree some tests are in order will cover in seprate PR.

stevewinsladeukef (approved):

ttbarnes (approved):

PR #1089, opened 2024-12-06
feat(APIM-609): companies house - extend returned fields

ttbarnes:
## Introduction :pencil2:
In order for Companies House to be consumed by other products/services, the integration needs to return some additional fields.

Note: unfortunately, some methods in the Companies House service do not have unit test coverage as they are private methods. I would like to add unit tests for all private methods (likely needing to make them public or extracted into helper functions), but this would be a bigger change than introducing additional fields.

- Update Companies House types.
- Update `CompaniesService` to return the following additional data:
  - `accounts` object
  - `dateOfCreation`
  - `registeredAddress.careOf`
  - `registeredAddress.premises`
  - `registeredAddress.region`
- Updated test generator shapes.

# Todo ❗ 
* Ensure tests are DRY and centralised helpers for types, DTO and interfaces.


avaitonis (approved):
Looks good

PR #1112, opened 2024-12-18
feat(GIFT-6804): setup connection to ods

ChristianMcc:
## Introduction :pencil2:
We need to be able to integrate with ODS to fetch data for GIFT

Add new config and database module for ODS, keeping it as similar as the other module files but using a service principal to connect to the DB instead.
I've tested this on an endpoint and it was happy. Tests aren't running as the PR isn't against main (I'll PR my FN-3543 to main in January later on) but they are all passing locally


docker-compose.yml
```diff
@@ -28,6 +28,11 @@ services:
       DATABASE_NUMBER_GENERATOR_NAME:
       DATABASE_CEDAR_NAME:
       DATABASE_CIS_NAME:
+      DATABASE_ODS_HOST:
```
ChristianMcc, line 31: I've actively chosen to include ODS_HOST separately to the other DATABASE_HOST so that we have the ability to point to different ODS instances (e.g. in staging where we connect to TEST for the CIS/CEDAR databases we might want to connect to PRE-PROD ODS)

src/modules/database/mssql-ods-database.module.ts
```diff
@@ -0,0 +1,37 @@
...
+          options: {
+            clientId: configService.get<string>('database.mssql_ods.client_id'),
+            clientSecret: configService.get<string>('database.mssql_ods.client_secret'),
+            tenantId: configService.get<string>('database.mssql_ods.tenant_id'),
+          },
+        },
+        extra: {
+          options: {
+            encrypt: true,
+            trustServerCertificate: true,
```
avaitonis, line 29: `trustServerCertificate: true` introduces security issue. We could ask IT to `whitelist` database.mssql_ods.host in our firewall. Then Azure sertificate should work fine. 

This can be TODO comment for later.

ChristianMcc: Yeah that's a good point, we should look at fixing all 4 of the database modules at once when we tackle this. I'll make a ticket in GIFT backlog for this so it's tracked (but might look at swapping that to APIM if we get an APIM board/wherever it makes sense to track this)

avaitonis: Good point about other databases.
APIM has Jira board for BAU, but it's not very tidy - https://ukef-dtfs.atlassian.net/jira/software/c/projects/APIM/boards/29
GIFT ticket might be fine.

ChristianMcc: Cheers, have made it on GIFT for now https://dev.azure.com/ukefprojects/GIFT/_workitems/edit/7414

src/modules/database/mssql-ods-database.module.ts
```diff
@@ -0,0 +1,37 @@
...
+    TypeOrmModule.forRootAsync({
+      name: DATABASE.ODS,
+      imports: [ConfigModule],
+      inject: [ConfigService],
+      useFactory: (configService: ConfigService) => ({
+        name: DATABASE.ODS,
+        host: configService.get<string>('database.mssql_ods.host'),
+        port: configService.get<number>('database.mssql_ods.port'),
+        database: configService.get<string>('database.mssql_ods.name'),
+        type: 'mssql',
```
abhi-markan, line 17: We have this repeated quite often, can you please convert this to a constant?

src/modules/database/mssql-ods-database.module.ts
abhi-markan: Can you follow CCM approach on this so it only impacts localhost only and not Azure enviornments.

abhi-markan (changes requested):
Nice work.

src/modules/database/mssql-ods-database.module.ts
ChristianMcc: I've extracted it, a lot of files are now changed due to renaming constants

src/modules/database/mssql-ods-database.module.ts
ChristianMcc: I need this on azure so I can't toggle it on/off for different environments

abhi-markan: Understood, will we then require `trustServerCertificate` property for Azure?

src/modules/database/mssql-ods-database.module.ts
abhi-markan: Thank you

abhi-markan (changes requested):

src/modules/database/mssql-ods-database.module.ts
ChristianMcc: yes that's right

PR #1127, opened 2025-01-03
feat(GIFT-3546): add endpoint to fetch customer data

ChristianMcc:
## Introduction :pencil2:
Create an endpoint and new ODS service/controller/module to query ODS for the customer.
Added types and functionality for new endpoints for other ODS entities to be fetched in the future.

All entities from ODS will be fetched in a similar way using the same stored procedure hence lumping it all together in a single ODS service, we could look at splitting up controllers and services in the future if we feel they are getting too bloated.

No full payload validation has been done of the response from the ODS stored procedures - only validation of the fields that is required for the endpoint.

An API test will be added separately to this PR in a following PR


src/modules/ods/dto/get-ods-customer-response.dto.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class GetOdsCustomerResponse {
+  @ApiProperty({
+    description: 'The unique UKEF id of the customer',
+  })
+  readonly customerUrn: string;
```
ChristianMcc, line 7: ditto on customerUrn comment on the other dto

src/modules/ods/ods.controller.test.ts
```diff
@@ -0,0 +1,31 @@
+import { CUSTOMERS } from '@ukef/constants';
+import { when } from 'jest-when';
+
+import { OdsController } from './ods.controller';
+import { OdsService } from './ods.service';
+
+describe('OdsController', () => {
```
ChristianMcc, line 7: I wanted a test to check the validation of the dto (e.g. valid when a urn matching the regex is passed through) but that would require setting up a validationPipe for this test which no other test in the codebase has done. For now I'm going to leave out to stay consistent with the other test suites but one to look at across the board

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,90 @@
...
+import { InjectDataSource } from '@nestjs/typeorm';
+import { DATABASE } from '@ukef/constants';
+import { DataSource } from 'typeorm';
+import { PinoLogger } from 'nestjs-pino';
+import { ODS_ENTITIES, OdsEntity, odsStoredProcedureInput, odsStoredProcedureQueryParams } from './dto/ods-payloads.dto';
+
+@Injectable()
+export class OdsService {
+  constructor(
+    @InjectDataSource(DATABASE.ODS)
```
ChristianMcc, line 12: This black box magic of adding the name of the data source to pick the right one is amazing, I don't fully understand it but I like it

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,90 @@
...
+      if (storedProcedureJson?.total_result_count == 0) {
+        throw new NotFoundException('No matching customer found');
+      }
+
+      return {
+        customerUrn: storedProcedureJson.results[0]?.customer_party_unique_reference_number,
+        customerName: storedProcedureJson.results[0]?.customer_name,
+      };
+    } catch (err) {
+      if (err instanceof NotFoundException) {
```
ChristianMcc, line 39: the catching and rethrowing feels a little clunky here but is the best way to get better logging for debugging which I think is worth it. Any recommendations on refactoring this are welcome

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,90 @@
...
+   * Calls the ODS stored procedure with the input provided and returns the output of it
+   * @param {odsStoredProcedureInput} storedProcedureInput The input parameter of the stored procedure
+   *
+   * @returns {Promise<any>} The result of the stored procedure
+   */
+  async callOdsStoredProcedure(storedProcedureInput: odsStoredProcedureInput): Promise<any> {
+    const queryRunner = this.odsDataSource.createQueryRunner();
+    try {
+      // Use the query runner to call a stored procedure
+      const result = await queryRunner.query(
```
ChristianMcc, line 76: Unfortuantely typeorm doesn't support any calling of stored procedures outside of just a SQL query so I've written a parameterised SQL query to stop SQL injection but has to be hard coded like this

src/modules/ods/dto/get-ods-customer-query.dto.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class GetOdsCustomerQueryDto {
+  @ApiProperty({
+    required: true,
+    example: CUSTOMERS.EXAMPLES.PARTYURN,
+    description: 'The unique UKEF id of the customer to search for.',
+    pattern: regexToString(UKEFID.PARTY_ID.REGEX),
+  })
+  @Matches(UKEFID.PARTY_ID.REGEX)
+  public customerUrn: string;
```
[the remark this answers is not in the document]
ChristianMcc: Opted to just call it urn while in the namespace of customer

src/modules/ods/ods.controller.test.ts
avaitonis: At the moment DTO validations are tested as part of `api/e2e` tests (`root/tests`). Most likely these tests are not part of github pipelines yet.

Would be nice to create `api/e2e` test for `osd`.

ChristianMcc: Yep will add this in a new PR straight after this one (don't have a tonne of time today/tomorrow but keen to get my feature branch tested in dev so have split up the work)

src/modules/ods/ods.service.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    } as unknown as jest.Mocked<QueryRunner>;
+
+    mockDataSource = {
+      createQueryRunner: jest.fn().mockReturnValue(mockQueryRunner),
+    } as unknown as jest.Mocked<DataSource>;
+
+    service = new OdsService(mockDataSource, null);
+  });
+
+  it('callOdsPrcedure should call the stored procedure with the query runner', async () => {
```
toddym42, line 24: `it` tests should generally read as a sentence starting with `it`. e.g.
```
it('does a thing', async () => { ...
```

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    };
+  }
+
+  /**
+   * Calls the ODS stored procedure with the input provided and returns the output of it
+   * @param {odsStoredProcedureInput} storedProcedureInput The input parameter of the stored procedure
+   *
+   * @returns {Promise<any>} The result of the stored procedure
+   */
+  async callOdsStoredProcedure(storedProcedureInput: odsStoredProcedureInput): Promise<any> {
```
toddym42, line 80: I think we should be able to give a type for the output, maybe an array of something like this?
```
output_body: {
  query_request_id: string;
  message: string;
  status: 'SUCCESS' | 'ERROR';
  total_result_count: number;
  results: Record<string, unknown>;
}
```
or if not sure if this is definitely returned could make the values optional?

Could link to [here](https://github.com/UK-Export-Finance/database-ods-datateam/blob/dev/t_apim/Stored%20Procedures/sp_ODS_query.sql#L279-L286) if we do want to store this type somewhere.

Could go further and make this a generic with the expected the type of `results` based on the particular query were running.

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,98 @@
...
+      const storedProcedureResult = await this.callOdsStoredProcedure(spInput);
+
+      const storedProcedureJson = JSON.parse(storedProcedureResult[0]?.output_body);
+
+      if (storedProcedureJson == undefined || storedProcedureJson?.status != 'SUCCESS') {
+        this.logger.error('Error from ODS stored procedure, output: %o', storedProcedureResult);
+        throw new InternalServerErrorException('Error trying to find a customer');
+      }
+
+      if (storedProcedureJson?.total_result_count == 0) {
```
toddym42, line 38: Any reason to use `==` here instead of `===`?

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,98 @@
...
+        DECLARE @output_body NVARCHAR(MAX);
+        EXEC t_apim.sp_ODS_query @input_body=@0, @output_body=@output_body OUTPUT;
+        SELECT @output_body as output_body
+      `,
+        [JSON.stringify(storedProcedureInput)],
+      );
+
+      return result;
+    } finally {
+      queryRunner.release();
```
toddym42, line 95: I think this needs an `await`?

src/modules/ods/ods.controller.test.ts
```diff
@@ -0,0 +1,31 @@
+import { CUSTOMERS } from '@ukef/constants';
+import { when } from 'jest-when';
```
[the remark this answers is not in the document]
ChristianMcc: There is unfortunately a package clash for this with the peer dependencies in this project which I'm going to call out of scope for this :(

src/modules/ods/ods.service.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      });
+
+      const mockResult = [{ output_body: JSON.stringify({ id: '123', name: 'Test Customer' }) }];
+      mockQueryRunner.query.mockResolvedValue(mockResult);
+
+      const result = await service.callOdsStoredProcedure(mockInput);
+
+      expect(mockDataSource.createQueryRunner).toHaveBeenCalled();
+      expect(mockQueryRunner.query).toHaveBeenCalledWith(
+        expect.stringMatching(
```
ChristianMcc, line 37: I didn't want to have to match indentation in the string matching so there is a regex match instead. Thoughts @toddym42 ?

src/modules/ods/ods.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  @Get('customers/:customerUrn')
+  @ApiOperation({
+    summary: 'Get customers from ODS',
+  })
+  @ApiResponse({
+    status: 200,
+    description: 'Customers matching search parameters',
+    type: GetOdsCustomerResponse,
+  })
```
toddym42, line 20: Not requesting to change, but just an fyi that there is a more specific `@ApiOkResponse` that we could use here, like with the error responses.

src/modules/ods/ods.service.test.ts
toddym42: Not sure I can think of anything better if we want to match the whole string, but I'm wondering if there's much value asserting on the whole thing, as it's just a hard-coded value that we're repeating. Could just expect any string, or something more minimal like `expect.stringContaining('EXEC t_apim.sp_ODS_query')`?

src/modules/ods/ods.service.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        ),
+        [JSON.stringify(mockInput)],
+      );
+      expect(result).toEqual(mockResult);
+      expect(mockQueryRunner.release).toHaveBeenCalled();
+    });
+  });
+
+  describe('findCustomer', () => {
+    it('findCustomer should return a customer the customer urn and name when findCustomer is called', async () => {
```
toddym42, line 48:
```suggestion
    it('should return a customer, the customer urn, and name when findCustomer is called', async () => {
```

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+   */
+  async findCustomer(urn: string): Promise<GetOdsCustomerResponse> {
+    try {
+      const spInput = this.createOdsStoredProcedureInput(ODS_ENTITIES.CUSTOMER, { customer_party_unique_reference_number: urn });
+
+      const storedProcedureResult = await this.callOdsStoredProcedure(spInput);
+
+      const storedProcedureJson: OdsStoredProcedureOuputBody = JSON.parse(storedProcedureResult[0]?.output_body);
+
+      if (storedProcedureJson === undefined || storedProcedureJson?.status != 'SUCCESS') {
```
toddym42, line 40:
```suggestion
      if (storedProcedureJson === undefined || storedProcedureJson?.status !== 'SUCCESS') {
```
(sorry I missed this first time round 😅)

toddym42 (approved):

src/modules/ods/ods.service.ts
abhi-markan: Can it not just be `storedProcedureJson?.status !== 'SUCCESS'`, is there a reason for an explicity undefined comparison?

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.logger.error('Error from ODS stored procedure, output: %o', storedProcedureResult);
+        throw new InternalServerErrorException('Error trying to find a customer');
+      }
+
+      if (storedProcedureJson?.total_result_count === 0) {
+        throw new NotFoundException('No matching customer found');
+      }
+
+      return {
+        urn: storedProcedureJson.results[0]?.customer_party_unique_reference_number,
```
abhi-markan, line 50: Do we always expect `results` to be an array?

src/modules/ods/ods.service.ts
abhi-markan: Its not ideal but to be honest, I can see reminisense of this pattern elsewhere in the codebase. Ideally an exceptional class will be clean but might creep the scope of this PR.

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Calls the ODS stored procedure with the input provided and returns the output of it
+   * @param {OdsStoredProcedureInput} storedProcedureInput The input parameter of the stored procedure
+   *
+   * @returns {Promise<OdsStoredProcedureOuput>} The result of the stored procedure
+   */
+  async callOdsStoredProcedure(storedProcedureInput: OdsStoredProcedureInput): Promise<OdsStoredProcedureOutput> {
+    const queryRunner = this.odsDataSource.createQueryRunner();
+    try {
```
abhi-markan, line 89: Do you not need `await queryRunner.connect();`?

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+      const result = await queryRunner.query(
+        `
+        DECLARE @output_body NVARCHAR(MAX);
+        EXEC t_apim.sp_ODS_query @input_body=@0, @output_body=@output_body OUTPUT;
+        SELECT @output_body as output_body
+      `,
+        [JSON.stringify(storedProcedureInput)],
+      );
+
+      return result;
```
abhi-markan, line 100:
```suggestion
 // Return the data or falsy
 return result[0]?.output_body || null;
```

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    } finally {
```
abhi-markan, line 101: Missed catch block.

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    const queryRunner = this.odsDataSource.createQueryRunner();
+    try {
+      // Use the query runner to call a stored procedure
+      const result = await queryRunner.query(
+        `
+        DECLARE @output_body NVARCHAR(MAX);
+        EXEC t_apim.sp_ODS_query @input_body=@0, @output_body=@output_body OUTPUT;
+        SELECT @output_body as output_body
+      `,
+        [JSON.stringify(storedProcedureInput)],
```
abhi-markan, line 97: * Is `storedProcedureInput` validated before passing it to the query?
* Does SP has enought validation to ensure no malcious code wil be executed.
* Does SP has correct ACL i.e. role management?
* Error handling should also be sanitized to ensure error messages do not expose anything sensitive back to the client.

abhi-markan (changes requested):
Nice work, few minor and medium comments to address please.

src/modules/ods/ods.service.ts
ChristianMcc: It was there for code readability showing that it can be undefined rather than any required programmatical check, I'm happy to remove

src/modules/ods/dto/ods-payloads.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+export type OdsStoredProcedureOuputBody = {
+  query_request_id: string;
+  message: string;
+  status: 'SUCCESS' | 'ERROR';
+  total_result_count: number;
+  results: Record<string, any>[];
+};
+
+export const ODS_ENTITIES = {
+  CUSTOMER: 'customer',
```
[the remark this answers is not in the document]
ChristianMcc: This is a constant I've defined to have all the different entities that can be queried in ODS

src/modules/ods/ods.service.ts
ChristianMcc: good catch!

src/modules/ods/ods.service.ts
ChristianMcc: I didnt' have a catch block here as it'll be caught in the function above - I can add one which will just be:

```
catch (error) {
  throw error
}
```
It won't add or do anything but if it's a pattern we want to stick to will add it

src/modules/ods/ods.service.ts
ChristianMcc: - we create storedProcedureInput from code and from a validated input to the endpoint (urn).
- we use parameterisation to stop SQL injection
- The user (service principal) calling this stored procedure has a very tight ACL, it can only run this stored procedure and query one logs table, it can't do anything else
- None of our error handling uses the output of the stored procedure, we only log that to Azure server logs

src/modules/ods/dto/get-ods-customer-param.dto.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class GetOdsCustomerParamDto {
+  @ApiProperty({
+    required: true,
+    example: CUSTOMERS.EXAMPLES.PARTYURN,
+    description: 'The unique UKEF id of the customer to search for.',
+    pattern: regexToString(UKEFID.PARTY_ID.REGEX),
+  })
+  @Matches(UKEFID.PARTY_ID.REGEX)
+  public urn: string;
```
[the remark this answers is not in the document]
ChristianMcc: Because a lot of urns start with 00 we don't want to cast this to number, has to stay a string

src/modules/ods/dto/get-ods-customer-response.dto.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class GetOdsCustomerResponse {
+  @ApiProperty({
+    description: 'The unique UKEF id of the customer',
+    example: '00312345',
+  })
+  readonly urn: string;
```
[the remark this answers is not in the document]
ChristianMcc: ditto other comment: Because a lot of urns start with 00 we don't want to cast this to number, has to stay a string

src/modules/ods/ods.service.ts
ChristianMcc: Nope the `createQueryRunner` function is synchronous :)

src/modules/ods/ods.service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Calls the ODS stored procedure with the input provided and returns the output of it
+   * @param {OdsStoredProcedureInput} storedProcedureInput The input parameter of the stored procedure
+   *
+   * @returns {Promise<OdsStoredProcedureOuput>} The result of the stored procedure
+   */
+  async callOdsStoredProcedure(storedProcedureInput: OdsStoredProcedureInput): Promise<string> {
+    const queryRunner = this.odsDataSource.createQueryRunner();
+    try {
```
abhi-markan, line 88:
```suggestion

    try {
```

src/modules/ods/ods.service.ts
abhi-markan: If we can please to keep the code clean that will be great. Thanks

abhi-markan (changes requested):
Looks good, two minor comments please. Thanks

abhi-markan (approved):

PR #1163, opened 2025-01-21
feat(FN-3664): add risk data to created customers

natdeanlewissoftwire:
## Introduction :pencil2:
In discussion with Risk teams, they wanted Risk data to be added to newly created customers in Salesforce.
The APIM side will default Risk rating at B+ and Loss given default at 50%, and the DTFS side will pass Probability of Default through to APIM to be set on the new record.

Defaults the first two risk values, and sets the Probability of Default value to the value DTFS passes through.

A small doc typo fix I found.

### DIFF
--- a/src/helpers/date-formatter.helper.test.ts
+++ b/src/helpers/date-formatter.helper.test.ts
@@ -1,17 +1,22 @@
 import { salesforceFormattedCurrentDate } from './date-formatter.helper';
 
 describe('salesforceFormattedCurrentDate function', () => {
-  const mockDate = new Date(2007, 3, 27);
-
-  beforeEach(() => {
-    jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
-  });
-
   afterEach(() => {
     jest.restoreAllMocks();
   });
 
-  it('should return the current date, formatted correctly', () => {
-    expect(salesforceFormattedCurrentDate()).toBe('2007-04-27');
+  const testCases = [
+    { mockDate: new Date('2007-04-27T00:00:00Z'), expected: '2007-04-27' },
+    { mockDate: new Date('2007-04-27'), expected: '2007-04-27' },
+    { mockDate: new Date(2007, 3, 27), expected: '2007-04-27' },
+    { mockDate: new Date('1970-01-01T12:34:56Z'), expected: '1970-01-01' },
+    { mockDate: new Date('9999-12-31'), expected: '9999-12-31' },
+    { mockDate: new Date('2020-02-29T00:00:00Z'), expected: '2020-02-29' }, // Leap year
+  ];
+
+  test.each(testCases)('should format the date $input as $expected', ({ mockDate, expected }) => {
+    jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
+
+    expect(salesforceFormattedCurrentDate()).toBe(expected);
   });
 });
--- a/src/helpers/date-formatter.helper.test.ts
+++ b/src/helpers/date-formatter.helper.test.ts
@@ -1,0 +1,17 @@
+import { salesforceFormattedCurrentDate } from './date-formatter.helper';
+
+describe('salesforceFormattedCurrentDate function', () => {
+  const mockDate = new Date(2007, 3, 27);
+
+  beforeEach(() => {
+    jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should return the current date, formatted correctly', () => {
+    expect(salesforceFormattedCurrentDate()).toBe('2007-04-27');
+  });
+});
--- a/src/helpers/date-formatter.helper.ts
+++ b/src/helpers/date-formatter.helper.ts
@@ -9,5 +9,6 @@ export const salesforceFormattedCurrentDate = () => {
   const dd = String(today.getDate()).padStart(2, '0');
   const mm = String(today.getMonth() + 1).padStart(2, '0');
   const yyyy = String(today.getFullYear());
+
   return `${yyyy}-${mm}-${dd}`;
 };
--- a/src/helpers/date-formatter.helper.ts
+++ b/src/helpers/date-formatter.helper.ts
@@ -1,0 +1,13 @@
+/**
+ * Returns the current date in the correct format for ingestion by the Salesforce sObject API.
+ *
+ * @returns {string} the current date in yyyy-mm-dd format
+ */
+
+export const salesforceFormattedCurrentDate = () => {
+  const today = new Date();
+  const dd = String(today.getDate()).padStart(2, '0');
+  const mm = String(today.getMonth() + 1).padStart(2, '0');
+  const yyyy = String(today.getFullYear());
+  return `${yyyy}-${mm}-${dd}`;
+};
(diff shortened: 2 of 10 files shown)


src/helpers/date-formatter.helper.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Returns the current date in the correct format for ingestion by the Salesforce sObject API.
```
abhi-markan, line 2:
```suggestion
 * Returns the current date in the correct format for ingestion by the Salesforce Object API.
```

src/helpers/date-formatter.helper.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  beforeEach(() => {
+    jest.spyOn(global, 'Date').mockImplementation(() => mockDate);
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should return the current date, formatted correctly', () => {
```
abhi-markan, line 14: Can we add more test cases:

1. Past date 1970-01-01.
2. Future date 9999-99-99.
3. Leap year date 2020-02-29.

src/helpers/date-formatter.helper.ts
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * @returns {string} the current date in yyyy-mm-dd format
+ */
+
+export const salesforceFormattedCurrentDate = () => {
+  const today = new Date();
+  const dd = String(today.getDate()).padStart(2, '0');
+  const mm = String(today.getMonth() + 1).padStart(2, '0');
+  const yyyy = String(today.getFullYear());
+  return `${yyyy}-${mm}-${dd}`;
```
abhi-markan, line 12:
```suggestion
  
  return `${yyyy}-${mm}-${dd}`;
```

src/modules/customers/customers.service.ts
```diff
@@ -205,6 +206,10 @@ export class CustomersService {
       Party_URN__c: partyUrn,
       D_B_Number__c: dunsNumber,
       Company_Registration_Number__c: DTFSCustomerDto.companyRegistrationNumber,
+      CCM_Credit_Risk_Rating__c: 'B+',
```
abhi-markan, line 209: Can this be a constant please.

src/modules/customers/customers.service.ts
```diff
@@ -205,6 +206,10 @@ export class CustomersService {
...
+      CCM_Credit_Risk_Rating_Date__c: salesforceFormattedCurrentDate(),
+      CCM_Loss_Given_Default__c: 50,
```
abhi-markan, line 211: Can this be a constant please.

src/modules/customers/dto/create-customer.dto.ts
```diff
@@ -25,4 +25,28 @@ export class CreateCustomerDto {
...
   @MaxLength(10)
   Company_Registration_Number__c: string;
+
+  @ApiProperty({ description: 'Credit Risk Rating' })
+  @IsString()
+  @IsNotEmpty()
+  CCM_Credit_Risk_Rating__c: string;
+
+  @ApiProperty({ description: 'Credit Risk Rating Date' })
+  @IsString()
```
abhi-markan, line 35: Will this be in `YYYY-MM-DD` format, if yes can we set the length decorator to `10`?

src/modules/customers/dto/create-customer.dto.ts
```diff
@@ -25,4 +25,28 @@ export class CreateCustomerDto {
...
+  @IsNotEmpty()
+  CCM_Credit_Risk_Rating_Date__c: string;
+
+  @ApiProperty({ description: 'Loss Given Default' })
+  @IsNumber()
+  @IsNotEmpty()
+  @Min(0)
+  @Max(100)
```
abhi-markan, line 43: Is the maximum `100`, want to make sure this can also be a float number or a string with percentage sign?

src/modules/customers/dto/create-customer.dto.ts
```diff
@@ -25,4 +25,28 @@ export class CreateCustomerDto {
...
+  CCM_Loss_Given_Default__c: number;
+
+  @ApiProperty({ description: 'Probability of Default' })
+  @IsNumber()
+  @IsNotEmpty()
+  @Min(0)
+  @Max(100)
```
abhi-markan, line 50: * Is the maximum `100` or `100.00`?

src/modules/customers/dto/dtfs-customer.dto.ts
```diff
@@ -13,4 +13,9 @@ export class DTFSCustomerDto {
   @IsString()
   @IsNotEmpty()
   companyName: string;
+
+  @ApiProperty({ description: 'Probability of Default' })
+  @IsNumber()
+  @IsNotEmpty()
```
abhi-markan, line 19: Will this accept `14.1` as a default PoD from DTFS?

src/modules/salesforce/salesforce.service.test.ts
```diff
@@ -105,6 +109,10 @@ describe('SalesforceService', () => {
         Party_URN__c: null,
         D_B_Number__c: null,
         Company_Registration_Number__c: '12341234',
+        CCM_Credit_Risk_Rating_Date__c: '2007-03-27',
+        CCM_Credit_Risk_Rating__c: 'B+',
+        CCM_Loss_Given_Default__c: 50,
+        CCM_Probability_of_Default__c: null,
```
abhi-markan, line 115: Can we add some assertions with edge values.
PoD: 100
Risk Rating: AAA+
LGD: 100

test/docs/__snapshots__/get-docs-yaml.api-test.ts.snap
```diff
@@ -808,9 +808,13 @@ components:
         companyName:
           type: string
           description: Company Name
+        probabilityOfDefault:
+          type: number
+          description: Probability of Default
```
abhi-markan, line 813: Is this missing LGD and customer rating?

abhi-markan (changes requested):

src/helpers/date-formatter.helper.ts
natdeanlewissoftwire: Funnily enough this is actually not a typo - I think the s stands for salesforce?
https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_methods_system_sobject.htm

src/modules/customers/dto/create-customer.dto.ts
natdeanlewissoftwire: Sure, good shout

src/modules/customers/dto/dtfs-customer.dto.ts
natdeanlewissoftwire: Will add a test for this

test/docs/__snapshots__/get-docs-yaml.api-test.ts.snap
natdeanlewissoftwire: Ah thanks I'll regenerate

src/modules/customers/dto/create-customer.dto.ts
natdeanlewissoftwire: I'm enforcing this to be a float/number not a string which I think is reasonable. That's how it's stored in DTFS anyway so seems to be the most consistent choice. Floats pass these conditons as to plain integers.

src/modules/customers/dto/create-customer.dto.ts
natdeanlewissoftwire: see above, 100.00 passes this, 100 is fine too. Will add some tests for this behaviour.

PR #1182, opened 2025-01-28
chore(APIM-612): companies house - isactive field

ttbarnes:
## Introduction :pencil2:
In order for Companies House to be consumed by other products/services, the integration should return an `isActive` field.

This will ensure that any product/service that needs to check if a company is active, will not need to perform a check on the `company_status` field and instead, can simply consume a `isActive` field.

- Update the service to return a `isActive` field.
- Update types and `GetCompanyGenerator`.

### DIFF
--- a/src/modules/companies/companies.service.ts
+++ b/src/modules/companies/companies.service.ts
@@ -84,6 +84,7 @@ export class CompaniesService {
         region: address?.region,
       },
       industries: this.mapSicCodes(company.sic_codes, industryClasses),
+      isActive: company.company_status === 'active',
     };
   }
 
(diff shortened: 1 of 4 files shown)


test/support/generator/get-company-generator.ts
```diff
@@ -228,6 +229,7 @@ export class GetCompanyGenerator extends AbstractGenerator<CompanyValues, Genera
         country: v.country,
       },
       industries,
+      isActive: v.isActive,
```
abhi-markan, line 232: Will `isActive` always be providedm wondering if this should be optional chained?

src/modules/companies/companies.service.ts
```diff
@@ -84,6 +84,7 @@ export class CompaniesService {
         region: address?.region,
       },
       industries: this.mapSicCodes(company.sic_codes, industryClasses),
+      isActive: company.company_status === 'active',
```
abhi-markan, line 87: Can `active` please be added to constant.

abhi-markan (changes requested):

abhi-markan (approved):

PR #1201, opened 2025-02-06
feat(FN-3163): allow non gef deals

natdeanlewissoftwire:
## Introduction :pencil2:
From QA testing it was recognised that it would be good to allow non-GEF deals to automatically create customers in Salesforce (as well as GEF deals which already have this behaviour implemented). The only blocker to this is that Probability of Default needs to be optional as non-GEF exporters don't have a Probability of Default entered.

Make Probability of Default optional.

### DIFF
--- a/src/modules/customers/customers.controller.test.ts
+++ b/src/modules/customers/customers.controller.test.ts
@@ -117,7 +117,17 @@ describe('CustomersController', () => {
   });
 
   describe('getOrCreateCustomer', () => {
-    const DTFSCustomerDto: DTFSCustomerDto = { companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG, companyName: 'TEST NAME', probabilityOfDefault: 3 };
+    const DTFSCustomerDtoWithProbabilityOfDefault: DTFSCustomerDto = {
+      companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG,
+      companyName: 'TEST NAME',
+      probabilityOfDefault: 3,
+    };
+
+    const DTFSCustomerDtoWithoutProbabilityOfDefault: DTFSCustomerDto = {
+      companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG,
+      companyName: 'TEST NAME',
+      probabilityOfDefault: undefined,
+    };
 
     const getOrCreateCustomerResponse: GetCustomersResponse = [
       {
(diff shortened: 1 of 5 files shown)


src/modules/salesforce/salesforce.service.test.ts
```diff
@@ -83,6 +83,16 @@ describe('SalesforceService', () => {
...
       },
+      {
+        Name: companyRegNo,
+        Party_URN__c: '00312345',
+        D_B_Number__c: '12341234',
+        Company_Registration_Number__c: companyRegNo,
+        CCM_Credit_Risk_Rating_Date__c: '2007-03-27',
+        CCM_Credit_Risk_Rating__c: CUSTOMERS.EXAMPLES.CREDIT_RISK_RATING,
+        CCM_Loss_Given_Default__c: CUSTOMERS.EXAMPLES.LOSS_GIVEN_DEFAULT,
+        CCM_Probability_of_Default__c: undefined,
```
abhi-markan, line 94: An assertion without `CCM_Probability_of_Default__c` please.

src/modules/customers/customers.controller.test.ts
```diff
@@ -117,7 +117,17 @@ describe('CustomersController', () => {
...
 
   describe('getOrCreateCustomer', () => {
-    const DTFSCustomerDto: DTFSCustomerDto = { companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG, companyName: 'TEST NAME', probabilityOfDefault: 3 };
+    const DTFSCustomerDtoWithProbabilityOfDefault: DTFSCustomerDto = {
+      companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG,
+      companyName: 'TEST NAME',
+      probabilityOfDefault: 3,
+    };
+
+    const DTFSCustomerDtoWithoutProbabilityOfDefault: DTFSCustomerDto = {
```
abhi-markan, line 126:
```suggestion
    const DTFSCustomerDtoWithFalsyProbabilityOfDefault: DTFSCustomerDto = {
```

src/modules/customers/customers.controller.test.ts
```diff
@@ -117,7 +117,17 @@ describe('CustomersController', () => {
...
+      companyRegistrationNumber: CUSTOMERS.EXAMPLES.COMPANYREG,
+      companyName: 'TEST NAME',
+      probabilityOfDefault: undefined,
+    };
```
abhi-markan, line 131: Can we add one without `probabilityOfDefault` field in the DTO.

abhi-markan (changes requested):

PR #1203, opened 2025-02-07
feat(GIFT-3546): deploy base 3546 branch to dev

ChristianMcc:
## Introduction :pencil2:
Enable deployments from feature branch to dev and small fix on path parameter and update with main


.github/workflows/deployment.yml
```diff
@@ -50,7 +51,7 @@ jobs:
     runs-on: [self-hosted, APIM, deployment]
     outputs:
       product: ${{ env.PRODUCT }}
-      environment: ${{ env.ENVIRONMENT }}
+      environment: ${{ github.ref_name == 'feat/FN-3543/integrate-with-ods' && 'dev' || env.ENVIRONMENT }}
```
abhi-markan, line 54:
```suggestion
      # TODO: Remove this branch name before merging to staging
      environment: ${{ github.ref_name == 'feat/FN-3543/integrate-with-ods' && 'dev' || env.ENVIRONMENT }}
```

src/modules/ods/dto/get-ods-customer-param.dto.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Matches } from 'class-validator';
+
+export class GetOdsCustomerParamDto {
+  @ApiProperty({
+    required: true,
+    example: CUSTOMERS.EXAMPLES.PARTYURN,
+    description: 'The unique UKEF ID of the customer to search for.',
+    pattern: regexToString(UKEFID.PARTY_ID.REGEX),
+  })
+  @Matches(UKEFID.PARTY_ID.REGEX)
```
abhi-markan, line 13: Can we please add `@min`, `@max` and/or `@length` decorators to keep it inline with CCM PRs.

src/modules/ods/dto/get-ods-customer-response.dto.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class GetOdsCustomerResponse {
+  @ApiProperty({
+    description: 'The unique UKEF ID of the customer',
+    example: '00312345',
+  })
+  readonly urn: string;
```
abhi-markan, line 8: Should we add same decorators to a response, might be worth checking CCM PRs.

src/modules/ods/ods.controller.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  beforeEach(() => {
+    const odsService = new OdsService(null, null);
+    odsServiceFindCustomer = jest.fn();
+    odsService.findCustomer = odsServiceFindCustomer;
+
+    controller = new OdsController(odsService);
+  });
+
+  describe('findCustomer', () => {
+    it('should return the customer when a valid customer URN is provided', async () => {
```
abhi-markan, line 20: Can we add an assertion with invalid and short URNs i.e. few unhappy paths please.

src/modules/ods/ods.service.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        ),
+        [JSON.stringify(mockInput)],
+      );
+      expect(result).toEqual(mockOutputBody);
+      expect(mockQueryRunner.release).toHaveBeenCalled();
+    });
+  });
+
+  describe('findCustomer', () => {
+    it('should return a customer the customer urn and name when findCustomer is called', async () => {
```
abhi-markan, line 49: Can we add an assertion where ODS does not return any response.
Bonus: connection failure to ODS please.

abhi-markan (changes requested):
Nice work, few comments to address please.

src/modules/ods/ods.controller.test.ts
ChristianMcc: all of these will be covered by the API tests as these are testing the decorators on the input and we don't do that anywhere else in mdm-api in the unit tests (it would require me setting up the validation pipeline within the unit test which would be a new pattern)

src/modules/ods/dto/get-ods-customer-response.dto.ts
ChristianMcc: Decorators from classValidator aren't used on responses (as we wouldn't be validating that) so leaving this

src/modules/ods/ods.service.test.ts
ChristianMcc: Have added one calling ODS throws which should cover a failed connection or ODS not returning us a response or timing out

abhi-markan (approved):

PR #1248, opened 2025-03-04
fix(file-upload): upload file as a buffer type

abhi-markan:
## Introduction :pencil2:
Ensure file upload is send to GovNotify as a `Buffer` type instead of `string` type.

* Introduced `convertStringToBuffer` helper to convert string to buffer.
* Updated test mock to sample buffer (CSV file).
* Removed maximum length for `file` property.
* Updated test cases.

### DIFF
--- a/src/helpers/convert-string-to-buffer.test.ts
+++ b/src/helpers/convert-string-to-buffer.test.ts
@@ -46,7 +46,7 @@ describe('convertStringToBuffer', () => {
     expect(result).toEqual(expectedBuffer);
   });
 
-  it('should return empty buffer for an empty string', () => {
+  it('should return an empty buffer for an empty string', () => {
     // Arrange
     const stringBuffer = '<Buffer>';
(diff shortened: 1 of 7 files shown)


src/helpers/convert-string-to-buffer.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    ]);
+
+    // Act
+    const result = convertStringToBuffer(stringBuffer);
+
+    // Assert
+    expect(result).toEqual(expectedBuffer);
+  });
+
+  it('should return empty buffer for an empty string', () => {
```
Zainzzkk, line 49: not a blocker, next PR
```suggestion
  it('should return an empty buffer for an empty string', () => {
```

Zainzzkk (dismissed):

Zainzzkk (dismissed):

Zainzzkk (approved):

